PR #8, opened 2021-10-25
fix: vite alias config

puppet-666:
**Description**
The vite alias parsing error occurs when the root path contains Chinese, but is correct when there is no Chinese
The file path is as follows
--项目
   --create-project
       --files
image


sodatea:
I can't reproduce and can't be sure the alias setting causes the error.

Next time, please open an issue with a minimal reproduction first. We need to make sure it's not a Vite bug.

PR #13, opened 2021-11-24
fix: use string as prop type in TS

cexbrayat:
The `HelloWorld` component was using `String` as a type instead of `string`.


sodatea (approved):
Thanks!

PR #14, opened 2021-12-01
fix: typo

btea:
fix comment typo


sodatea (approved):
Oops. I only manually wrote one line in this function and it contains a typo. Copilot clearly does a much better job than me 🤣

PR #17, opened 2021-12-03
fix(template): update gitignore to include extensions.json

jd-0001:
`.gitignore` had `.vscode`, with this config vscode ignores whole `.vscode` directory including `extensions.json`.

Updated to `.vscode/*` so it allows excluding files via `!` prefix.


sodatea (commented):
Thanks!

sodatea (approved):

PR #20, opened 2021-12-20
fix: remove default router ts

### LINKED ISSUE #19: Javascript project is created with lang="ts" in HomeView.vue
## Expected Behavior
First line of HomeView.vue:
`<script setup>`

## Current Behavior
First line of HomeView.vue:
`<script setup lang="ts">`

## Possible Solution
I can find the mistake inside `playground/router-pinia-with-tests/src/views/HomeView.vue` but I'm not sure which template was used to generate it.
All projects without typescript should be created without `lang="ts"` in any of the files.

## Steps to Reproduce
Scaffold a new project with:
1. NO Typescript
2. NO JSX
3. YES VueRouter
4. YES Pinia
5. YES Cypress

btea:
fix #19

### DIFF
--- a/playground/jsx-router-pinia-with-tests/src/views/HomeView.vue
+++ b/playground/jsx-router-pinia-with-tests/src/views/HomeView.vue
@@ -1,4 +1,4 @@
-<script setup lang="ts">
+<script setup>
 import TheWelcome from '@/components/TheWelcome.vue'
 </script>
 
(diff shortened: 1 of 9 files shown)


nicosemp:
Thank you! I wasn't sure if JSX files had to be included in the fix.

sodatea (approved):
Thanks!

PR #25, opened 2022-01-06
fix: eslint with prettier

cexbrayat:
The eslint script currently ignores the typescript, test, and prettier flags.


sodatea (approved):
😂 Thanks!

PR #26, opened 2022-01-06
feat: drop .vue shim and recommend typescript vue plugin by default

sodatea:
After discussions with @johnsoncodehk and @xiaoxiangmoe, we concluded that it makes more sense for most users to start with the TypeScript Vue Plugin instead of the `.vue` shim.

Take Over Mode is more performant and we would like to recommend it, but turning it on in VSCode takes too many steps. So I chose to only mention it in the README, so that more senior developers can know more about it.


NWYLZW:
WebStorm？

PR #34, opened 2022-01-21
chore: add `declare module "*.vue"` to "env.d.ts"

yaegassy:
Added `declare module "*.vue"` as in create-vite/template-vue-ts in vite. <https://github.com/vitejs/vite/blob/main/packages/create-vite/template-vue-ts/src/env.d.ts>

If there are any problems, please close this PR. 🙇

### DIFF
--- a/template/config/typescript/env.d.ts
+++ b/template/config/typescript/env.d.ts
@@ -1,1 +1,8 @@
 /// <reference types="vite/client" />
+
+declare module '*.vue' {
+  import { DefineComponent } from 'vue'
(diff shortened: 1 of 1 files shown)


yaegassy:
I checked this. <https://github.com/vuejs/create-vue/pull/26> 

I will close my PR.

---

**Supplement**:

I am using the plugin for volar from Vim. The `TypeScript Vue Plugin (Volar)` is not available outside of VSCode.

Also, unfortunately, I think there are some IDE/Editor plugins that do not support volar's Take OverMode.

Fortunately, the volar plugin for Vim that I use supports Take Over Mode.

PR #37, opened 2022-01-23
feat: add format command by prettier

tetradice:
If project is configured to use Prettier, configure it to be able to call prettier with the `npm run format` command.


sodatea:
You don't need a separate `format` command as `eslint-plugin-prettier` has already been included in the ESLint configuration. Just use `npm run lint`.

PR #39, opened 2022-01-24
fix: use vite dev during e2e dev

Akryum:


sodatea:
This makes some sense as all browsers that Cypress supports are modern browsers.
🤔 But on the other hand, unlike webpack, `vite dev` and `vite build` outputs differ greatly. Unit tests can't cover these differences.
So I would prefer to use E2E tests to make sure the build works correctly. Speed isn't much of a concern here.

sodatea:
Superseded by https://github.com/vuejs/create-vue/pull/183
We now have a `test:e2e:dev` command to use the dev server and run the tests interactively.

PR #40, opened 2022-01-28
feat: use solution-style tsconfigs for more accurate type checking

sodatea:


template/tsconfig/base/tsconfig.json
```diff
@@ -0,0 +1,16 @@
+{
+  "extends": "@vue/tsconfig/tsconfig.web.json",
+  "include": ["env.d.ts", "src/**/*", "src/**/*.vue"],
+  "compilerOptions": {
+    "baseUrl": ".",
```
sodatea, line 5: `baseUrl` here is redundant because TypeScript >= 4.1 supports `paths` without `baseUrl`: https://devblogs.microsoft.com/typescript/announcing-typescript-4-1/#paths-without-baseurl

But if it presents, `paths` aliases will be resolved relative to `baseUrl`. So, to prevent users from accidentally adding an incorrect `baseUrl`, we add a dummy one here.

index.js
```diff
@@ -286,8 +286,18 @@ async function init() {
   }
   if (needsTypeScript) {
     render('config/typescript')
+
+    // Render tsconfigs
+    render('tsconfig/base')
+    if (needsCypress) {
+      render('tsconfig/cypress')
+    }
+    if (needsVitest) {
```
sodatea, line 295: Only the Vitest template needs an additional tsconfig, because it runs in Node.js with JSDOM, which is different than other files under `src/`.

Cypress Component Testing runs in an actual browser, and the Cypress type definitions will be brought in by `import { mount } from '@cypress/vue'`, so there's no need for a standalone tsconfig.

template/tsconfig/vitest/tsconfig.vitest.json
```diff
@@ -0,0 +1,8 @@
+{
+  "extends": "@vue/tsconfig/tsconfig.node.json",
+  "include": ["src/**/__tests__/*"],
+  "compilerOptions": {
+    "composite": true,
+    "types": ["node", "jsdom"]
```
sodatea, line 6: I intentionally didn't include the Vitest types (which injects globals such as `describe`, `it`), because explicit is better than implicit.

DannyFeliz:
Thank you for adding these comments explaining some decisions that were made

PR #44, opened 2022-02-07
Use '@/components' relative path instead of './'

martinszeltins:
Since we already have `/@` path shorthand set up, let's use that instead of `./`


sodatea:
`@/` is a *shorthand*. This change makes the paths even longer.

PR #45, opened 2022-02-07
Use relative path using '@/` instead of '../'

martinszeltins:
Since we already have `/@` path shorthand set up, let's use that instead of `./`


sodatea:
We don't need `@/` everywhere.
It's just a shorthand notation and requires more cognitive load for people to understand.

So I would recommend using it only if it truly improves readability - that is, when there are too many `../`s in the path that you can't count it in a glance, or when it shortens the path by a lot.

`@/` isn't much an improvement over `../` IMO. It is shorter by one character but is harder to understand.

PR #47, opened 2022-02-08
fix(vitest): allow to typecheck tests

cexbrayat:


template/tsconfig/vitest/package.json
```diff
@@ -1,4 +1,7 @@
 {
+  "scripts": {
+    "typecheck": "vue-tsc --noEmit -p tsconfig.vitest.json --composite false"
```
sodatea, line 3: This would overwrite the one in the typescript template: https://github.com/vuejs/create-vue/blob/d8723fb59c34d3e5599442903170240417efb47c/template/config/typescript/package.json#L4 and only type-check the specs.
I think we need a separate script or merge it with the default `typecheck` script or the `test:unit` script

sodatea (changes requested):

template/tsconfig/vitest/package.json
cexbrayat: What about `    "typecheck": "vue-tsc --noEmit && vue-tsc --noEmit -p tsconfig.vitest.json --composite false"` then?

sodatea: Looks good to me 👍

cexbrayat: PR updated!

sodatea (approved):

PR #51, opened 2022-02-11
fix: fix packageName when npm init vue set valid targetDir

chenpeng991105:
Fix package.json dosen't generate `name` when npm init vue set valid targetDir


sodatea (approved):
Thanks!

PR #57, opened 2022-02-15
fix(template): removed old volar extension

jd-0001:
Removed "johnsoncodehk.vscode-typescript-vue-plugin" for extension recommendation.

As per [this](https://discord.com/channels/793943652350427136/793943652350427139/894949633267232778) chat, I think "johnsoncodehk.vscode-typescript-vue-plugin" extension is no longer needed.

Am I right, @johnsoncodehk?


xiaoxiangmoe:
See https://github.com/vuejs/create-vue/pull/26

For most users, the configuration of Take Over Mode is too complicated. 
So for them, we recommend using `johnsoncodehk.vscode-typescript-vue-plugin`.

PR #56, opened 2022-02-15
fix: vitest tsconfig should include all files under `src`

sodatea:
Credit to @xiaoxiangmoe 

Fixes #55

Fixes the case that a `.spec.ts` file inside `__test__/` imports a
module outside of `__test__/` (a cross-(typescript-)project reference).

Note that VSCode language service looks for `include` patterns in `references`
top-down, so `tsconfig.app.json` must come before `tsconfig.vitest.json`
so that the `src/**` modules can get the most-accurate type hints.

The previous tsconfig only works for `.vue` imports in `.spec.ts`, which
seems to be a Volar bug. We shouldn't rely on that.
This fix is a more accurate configuration.

Vitest TypeScript projects created prior to `create-vue` 3.1.6 can apply the following patch to the projects:

```patch
From a37c33facf0e5db767bee33a71ad9e9fcd3cf373 Mon Sep 17 00:00:00 2001
From: Haoqun Jiang <haoqunjiang@gmail.com>
Date: Tue, 15 Feb 2022 21:50:07 +0800
Subject: [PATCH] fix: vitest tsconfig

---
 package.json         |  2 +-
 tsconfig.app.json    | 12 ++++++++++++
 tsconfig.json        | 14 ++++----------
 tsconfig.vitest.json |  9 +++------
 4 files changed, 20 insertions(+), 17 deletions(-)
 create mode 100644 tsconfig.app.json

diff --git a/package.json b/package.json
index ad9e16d..5138df6 100644
--- a/package.json
+++ b/package.json
@@ -6,7 +6,7 @@
     "build": "vue-tsc --noEmit && vite build",
     "preview": "vite preview --port 5050",
     "test:unit": "vitest --environment jsdom",
-    "typecheck": "vue-tsc --noEmit && vue-tsc --noEmit -p tsconfig.vitest.json --composite false",
+    "typecheck": "vue-tsc --noEmit -p tsconfig.vitest.json --composite false",
     "lint": "eslint . --ext .vue,.js,.jsx,.cjs,.mjs,.ts,.tsx,.cts,.mts --fix --ignore-path .gitignore"
   },
   "dependencies": {
diff --git a/tsconfig.app.json b/tsconfig.app.json
new file mode 100644
index 0000000..cdbea1d
--- /dev/null
+++ b/tsconfig.app.json
@@ -0,0 +1,12 @@
+{
+  "extends": "@vue/tsconfig/tsconfig.web.json",
[...]

PR #62, opened 2022-02-18
fix(cypress-ct): avoid cypress types polluting normal source files

sodatea:
As pointed out in https://github.com/vuejs/create-vue/pull/56#issuecomment-1040326589

### DIFF
--- a/template/tsconfig/cypress-ct/tsconfig.app.json
+++ b/template/tsconfig/cypress-ct/tsconfig.app.json
@@ -1,0 +1,12 @@
+{
+  "extends": "@vue/tsconfig/tsconfig.web.json",
+  "include": ["env.d.ts", "src/**/*", "src/**/*.vue"],
+  "exclude": ["src/**/__tests__/*"],
+  "compilerOptions": {
+    "composite": true,
+    "baseUrl": ".",
+    "paths": {
+      "@/*": ["./src/*"]
+    }
+  }
+}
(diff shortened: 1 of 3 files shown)


template/tsconfig/cypress-ct/tsconfig.app.json
```diff
@@ -0,0 +1,12 @@
+{
+  "extends": "@vue/tsconfig/tsconfig.web.json",
+  "include": ["env.d.ts", "src/**/*", "src/**/*.vue"],
```
xiaoxiangmoe, line 3: Maybe we need't "src/**/*.vue"

sodatea: I'm not sure about this. Even though `vue-tsc` doesn't need the `.vue` extension, many TypeScript-related tools are already assumed that only a particular group of "default extensions" are allowed.

For example, I just checked `tsconfck`, it has this assumption too:
https://github.com/dominikg/tsconfck/blob/a46f60c03c088dc2c6a1283b421919a398df6a05/src/util.ts#L174-L177

Vite uses `tsconfck` to get the corresponding typescript configuration for a given file path, because some of the tsconfig options will affect the compilation result (e.g. [preserveValueImports](https://www.typescriptlang.org/tsconfig/#preserveValueImports)).

So removing `.vue` here is a breaking change to those who relied on this behavior in `.vue` files.

---

Let's keep it as-is. After all, it's already a well-established convention in the TypeScript ecosystem.

xiaoxiangmoe (approved):

PR #71, opened 2022-02-24
fix: dont import json

zhmushan:
It would be cool to be able to execute projects directly using `node index.js`

### DIFF
--- a/index.js
+++ b/index.js
@@ -299,7 +299,8 @@ async function init() {
 
   // Render ESLint config
   if (needsEslint) {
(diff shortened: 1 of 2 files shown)


zhmushan:
@sodatea Any feedback?

sodatea:
Thanks for the PR.

But executing the code directly was never a goal. It doesn't help improve the project's maintainability.

I have now migrated the codebase to TypeScript, so importing JSON should be acceptable now.

PR #76, opened 2022-03-04
feat: add Playwright for end-to-end testing

mxschmitt:
Hello, 👋 
[Playwright](https://playwright.dev/) is getting more and more traction over the last years and we should provide users an easy integration with the Vue ecosystem so users benefit from it.
This Pull Request adds Playwright as an end-to-end testing solution like it's done for Cypress. It might make sense to throw an error if the user selects both variants for e2e testing.

We'll follow-up once the component testing story with Playwright and Vite is ready.

Thank you

cc @sodatea


template/config/playwright/playwright.config.ts
```diff
@@ -0,0 +1,103 @@
...
+  /* Shared settings for all the projects below. See https://playwright.dev/docs/api/class-testoptions. */
+  use: {
+    /* Maximum time each action such as `click()` can take. Defaults to 0 (no limit). */
+    actionTimeout: 0,
+    /* Base URL to use in actions like `await page.goto('/')`. */
+    baseURL: 'http://localhost:3001',
+
+    /* Collect trace when retrying the failed test. See https://playwright.dev/docs/trace-viewer */
+    trace: 'on-first-retry'
+  },
```
sodatea, line 41: How about setting the `headless` option based on the `process.env.CI` variable? Running in headed mode on local machines could provide a better experience IMO.

sodatea (changes requested):
The CI is failing. I think we need to also skip playwright tests in `scripts/test.mjs` since it requires different environment.

template/config/playwright/playwright.config.ts
```diff
@@ -0,0 +1,103 @@
...
+    // },
+  ],
+
+  /* Folder for test artifacts such as screenshots, videos, traces, etc. */
+  // outputDir: 'test-results/',
+
+  /* Run your local dev server before starting the tests */
+  webServer: {
+    command: 'npm run dev',
+    port: 3001
```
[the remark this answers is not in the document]
mxschmitt: Fixed, should have been 3000.

mxschmitt:
Applied the review feedback, sorry for the late response! Not sure if we should skip it inside the integration tests or run them - with my new implementation we run them.

(Would also be awesome if you could "lax" the GitHub Action permissions a big, so you don't have to approve all the time [here](https://github.com/vuejs/create-vue/settings/actions))

utils/generateReadme.ts
```diff
@@ -94,6 +95,23 @@ ${getCommand(packageManager, 'test:e2e')} # or \`${getCommand(
 `
   }
 
+  if (needsPlaywright) {
+    npmScriptsDescriptions += `
+### Run End-to-End Tests with [Playwright](https://playwright.dev)
+
+\`\`\`sh
+# Runs the end-to-end tests.
```
sodatea, line 103: Need to inform the user to run `npx playwright install` first.

utils/renderTemplate.ts
```diff
@@ -41,9 +41,13 @@ function renderTemplate(src, dest) {
     return
   }
 
-  if (filename.startsWith('_')) {
-    // rename `_file` to `.file`
-    dest = path.resolve(path.dirname(dest), filename.replace(/^_/, '.'))
+  if (filename === '.gitignore') {
```
sodatea, line 44: Please retain the renaming logic.
Because `.gitignore` won't be packed by npm, we have to use `_gitignore` in templates.

template/config/playwright/.github/workflows/playwright.yml
```diff
@@ -0,0 +1,27 @@
+name: Playwright Tests
```
sodatea, line 1: I feel that this file is unnecessary, because only a fraction of our users would use GitHub Actions.

template/config/vitest/package.json
```diff
@@ -1,6 +1,6 @@
 {
   "scripts": {
-    "test:unit": "vitest --environment jsdom"
+    "test:unit": "vitest --environment jsdom --root src/"
```
sodatea, line 3: Why changing this line?

index.ts
```diff
@@ -200,6 +203,14 @@ async function init() {
           active: 'Yes',
           inactive: 'No'
         },
+        {
+          name: 'needsPlaywright',
+          type: () => (isFeatureFlagsUsed ? null : 'toggle'),
```
sodatea, line 208: We should skip this prompt if the user has already chosen Cypress.

The implementation can be similar to the logic that `prettier` is skipped when ESLint hasn't been chosen: https://github.com/vuejs/create-vue/blob/604355fbbb0f5f29920a3cf5dedb24e50c9db623/index.ts#L213-L218

template/config/playwright/e2e/example.spec.ts
```diff
@@ -0,0 +1,398 @@
+import { test, expect, Page } from '@playwright/test';
+
+test.beforeEach(async ({ page }) => {
+  await page.goto('https://demo.playwright.dev/todomvc');
```
sodatea, line 4: This todomvc page is named "React" 😂
Do you have a Vue todomv demo?

Or maybe we can just remove the example spec? It's quite large and intended to be removed by the user anyway.

And we can put a link to the tutorial in `vue.spec.js/ts` in case the user don't know how to write specs.

sodatea (changes requested):
Sorry for the late review.

I've tried the PR locally and it's quite a joy to test it.
Just a few nitpicks that needs to be improved.

template/config/playwright/.github/workflows/playwright.yml
mxschmitt: I might be opinionated since MS -> GitHub but I think that the vast majority is using GitHub Actions nowadays.
Anyway, I removed it for now to have it less opinionated.

template/config/vitest/package.json
mxschmitt: Otherwise vitest tries to process the playwright test files as well, this won't work and lead to errors. Maybe there is a better way of handling that?

index.ts
mxschmitt: sweet!

template/config/playwright/e2e/example.spec.ts
mxschmitt: okay true, thats kinda weird if we kept it that way!
The main reason why we moved it to our domain is that we don't depend on the original mirror. All the todomvc sites are implementing the same application in different frontend frameworks, so it should work out of the box with the Vue one.
Will remove it for now.

mxschmitt:
> Sorry for the late review.
> 
> I've tried the PR locally and it's quite a joy to test it. Just a few nitpicks that needs to be improved.

No worries, was out of office myself so I did not notice at all. Applied the suggestions. Lets see what results the CI will give us.

stefnotch:
Out of curiosity, what's the current status on this? Playwright definitely is an interesting alternative to Cypress, so having it in there as an easy-to-access option seems lovely.

sand4rt:
@mxschmitt I wonder why this PR is closed?😕

mxschmitt:
@sodatea sorry for the huge delay, I was partially ooo back then and then I forgot about it.

Would be nice if you could enable GitHub Actions for contributors, this makes me easier to ensure its all green instead of having to wait for a maintainer to click approve.

I hope I addressed the comments, if not, please let me know!

The updates Playground PR is here:

sodatea (approved):
It looks good to me now.
Thank you for putting so much effort into this PR!

I'm merging the PR today.

But FYI, before cutting the next release, I'd like to make some adjustments to the workflow of this repo, the tsconfigs, and the CLI prompts, which would take some time.
So I expect this feature to be included in the October release.

template/config/vitest/package.json
ptrgast: Hello! It seems that this addition introduces a problem. When running Vitest it tries to create a `results.json` inside its cache directory. This directory is defined as `resolve(root, slash(dir || 'node_modules/.vitest'))`. This means that with `--root src/` added, Vitest creates the `results.json` inside the `<project_root>/src/node_modules/.vitest` instead of the actual node_modules directory (`<project_root>/node_modules/.vitest`).

PR #80, opened 2022-03-11
Add instructions for using Yarn

Robin-Hoodie:


sodatea:
`yarn create` doesn't allow the `@3` suffix. So it's not a good fit.

And I think those who are familiar with yarn must be smart enough to find out how to use this project with `yarn` by themselves. No need to bloat the documentation.

PR #83, opened 2022-04-10
re-add vue component type

silvercent011:


DannyFeliz:
I know that @sodatea will close this PR, same as me https://github.com/vuejs/create-vue/pull/65. I understand the explanation linked in my PR and I use the Take Over Mode in my projects but having a bunch of errors by default with a brand new project doesn't feel good, it would be better if we have this the module definition for vue file and just add some sort of recommendation about getting a better DX using the take over mode would be better approach IMHO.

angels7777 (commented):
THANKS

angels7777 (commented):
THANKS

xiaoxiangmoe:
@DannyFeliz @silvercent011 You can use [TypeScript Vue Plugin](https://marketplace.visualstudio.com/items?itemName=Vue.vscode-typescript-vue-plugin) rather than Take Over Mode.
I'm using TypeScript Vue Plugin now. And it works very well.

cexbrayat:
As mentionned in the comments above, this is no longer needed. Thanks!

PR #101, opened 2022-05-16
fix: tsconfig override by wrong reference order causing type missing.

miaulightouch:
the setting of `lib: []` in `tsconfig.vitest.json`, would cause the settings from `tsconfig.app.json` being cleanup. it would loss the type like `dom.interable`.


ebisbe:
I was wondering just what this PR is changing. My type-check was failing because it used tsconfig.vitest.json to build.

cexbrayat:
@miaulightouch Could you provide a repro showcasing the issue you're trying to fix?

 AFAIK, references should not override one another, and the order is not important, but maybe I'm missing something https://www.typescriptlang.org/docs/handbook/project-references.html

miaulightouch:
ok, it's weird. it's not shown error anymore after typescript updated both in old project and new created project.

PR #102, opened 2022-05-19
feat: add jsconfig.json for base template

GraxMonzo:
Create-vue doesn't create a `jsconfig.json` file on the first initialization without TS. Added `jsconfig.json` with aliased paths to base template for out of the box VSCode IntelliSense

### DIFF
--- a/template/base/jsconfig.json
+++ b/template/base/jsconfig.json
@@ -1,6 +1,5 @@
 {
   "compilerOptions": {
-    "baseUrl": ".",
     "paths": {
       "@/*": ["./src/*"]
     }
--- a/template/base/jsconfig.json
+++ b/template/base/jsconfig.json
@@ -1,0 +1,8 @@
+{
+  "compilerOptions": {
+    "baseUrl": ".",
+    "paths": {
+      "@/*": ["./src/*"]
+    }
+  }
+}
--- a/template/base/jsconfig.json
+++ b/template/base/jsconfig.json
@@ -3,5 +3,6 @@
     "paths": {
       "@/*": ["./src/*"]
     }
-  }
+  },
+  "exclude": ["node_modules", "dist"]
 }


cexbrayat (changes requested):
@GraxMonzo Thanks for the PR!

Have you tried it out? I'm wondering if putting the file in `base` won't add it for all types of projects, even the TS ones.

cexbrayat:
@sodatea Do you have an opinion on adding a `jsconfig.json` file in the generated (non-TS project)?

sodatea:
I'm fine with adding a `jsconfig.json`. But I'm not sure what are the best practices

GraxMonzo:
> @GraxMonzo Thanks for the PR!
> 
> Have you tried it out? I'm wondering if putting the file in `base` won't add it for all types of projects, even the TS ones.

Hey @cexbrayat. Yes I tried it and tried it without the `baseUrl` and it works. Well I guess `baseUrl` is redundant here, because as of TypeScript 4.1, you no longer need to specify `baseUrl` when using `paths`.
https://www.typescriptlang.org/tsconfig#baseUrl

Thank you for your time.

cexbrayat:
@GraxMonzo Sorry I wasn't clear: I meant that the file you added is in `template/base` so I'm wondering if it does not get added to all the generated projects with create-vue, even the one where the users pick TypeScript

GraxMonzo:
@cexbrayat Pardon me for misreading you. I haven't tested this with another template but I might think this line of code solves this problem https://github.com/vuejs/create-vue/blob/main/index.ts#L455

cexbrayat:
👍 Awesome, so maybe you can remove the baseUrl from the config if it isn't needed and then we'll be able to merge it

GraxMonzo:
@cexbrayat Done

sodatea (changes requested):
According to https://code.visualstudio.com/docs/languages/jsconfig#_examples we need to explicitly exclude `node_modules` and `dist` here.

sodatea (approved):

cexbrayat (approved):

cexbrayat:
Thanks @GraxMonzo !

PR #108, opened 2022-06-03
feat: update to Cypress 10

sodatea:


template/config/cypress-ct/cypress.config.js
```diff
@@ -0,0 +1,15 @@
+const { defineConfig } = require('cypress')
+
+module.exports = defineConfig({
+  e2e: {
+    specPattern: 'cypress/e2e/**/*.{cy,spec}.{js,jsx,ts,tsx}',
+    baseUrl: 'http://localhost:4173'
+  },
+  component: {
+    specPattern: 'src/**/__tests__/*.{cy,spec}.{js,ts,jsx,tsx}',
```
sodatea, line 9: I chose to explicitly specify `specPatterns` everywhere, for:
1. Make the convention clearer for newcomers;
2. Besides the default pattern, I also allowed `*.spec.*` so that it's easier for users from the old projects to migrate.

sodatea:
@JessicaSachs 
I pushed the templates to https://github.com/vuejs/create-vue-templates/tree/preview-cy-10

A basic example: https://github.com/vuejs/create-vue-templates/tree/preview-cy-10/cypress
Use with Vitest: https://github.com/vuejs/create-vue-templates/tree/preview-cy-10/vitest-cypress
Basic example, but in TypeScript: https://github.com/vuejs/create-vue-templates/tree/preview-cy-10/typescript-cypress
Use with Vitest, in TypeScript: https://github.com/vuejs/create-vue-templates/tree/preview-cy-10/typescript-vitest-cypress

template/config/cypress/package.json
```diff
@@ -1,10 +1,10 @@
 {
   "scripts": {
-    "test:e2e": "start-server-and-test preview http://127.0.0.1:4173/ 'cypress open'",
-    "test:e2e:ci": "start-server-and-test preview http://127.0.0.1:4173/ 'cypress run'"
+    "test:e2e": "start-server-and-test preview http://127.0.0.1:4173/ 'cypress open --e2e'",
```
sodatea, line 3: Currently `--e2e` and `--component` flags are always passed to skip the "Choose testing type" screen.
This is because the commands are `test:e2e` and `test:unit`, we expect users to know what types of tests they are looking for.

utils/renderEslint.ts
```diff
@@ -49,8 +49,8 @@ function configureEslint({ language, styleGuide, needsPrettier, needsCypress, ne
     const cypressOverrides = [
       {
         files: needsCypressCT
-          ? ['**/__tests__/*.spec.{js,ts,jsx,tsx}', 'cypress/integration/**.spec.{js,ts,jsx,tsx}']
-          : ['cypress/integration/**.spec.{js,ts,jsx,tsx}'],
+          ? ['**/__tests__/*.{cy,spec}.{js,ts,jsx,tsx}', 'cypress/e2e/**.{cy,spec}.{js,ts,jsx,tsx}']
+          : ['cypress/e2e/**.{cy,spec}.{js,ts,jsx,tsx}'],
```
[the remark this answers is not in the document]
JessicaSachs: I really prefer not using `__test__` but it's a conversation topic for another time.

`.spec.` will conflict with Vitest types I believe, which is why we chose `.cy.` but it's okay, you're right about the struggle people may have with migration.

template/config/cypress-ct/cypress.config.js
JessicaSachs: > Besides the default pattern, I also allowed *.spec.* so that it's easier for users from the old projects to migrate.

I believe this is okay for now.

utils/renderEslint.ts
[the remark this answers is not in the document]
sodatea: They will not in our case because of the usage of solution-style tsconfigs.
But yeah, people who are not familiar enough with TypeScript will find this troublesome. So I've changed the default pattern.

PR #109, opened 2022-06-03
fix: move Cypress links to "on" links so they do not break.

JessicaSachs:


sodatea (approved):
Didn't know this trick! Thanks!

PR #117, opened 2022-06-28
fix: resolve lint warnings in scaffold

### LINKED ISSUE #116: Scaffolded project has eslint(prettier/prettier) warnings
### Description

There is no `.pretterrc` in the scaffold, so Prettier uses its default options, which includes [`printWidth=80`](https://prettier.io/docs/en/options.html#print-width), [`semi=true`](https://prettier.io/docs/en/options.html#semicolons), and [`singleQuote=false`](https://prettier.io/docs/en/options.html#quotes). This default config causes Prettier warnings in almost all `.vue` files.

Also, the scaffolded `.eslintrc.cjs` contains double-quoted keys, which causes Prettier warnings:

```js
/* eslint-env node */
require("@rushstack/eslint-patch/modern-module-resolution"); // 💡 Replace `"@rushstack/eslint-patch/modern-module-resolution");` with `'@rushstack/eslint-patch/modern-module-resolution')`

module.exports = {
  "root": true, // 💡 Replace `"root"` with `root`
  "extends": [
    "plugin:vue/vue3-essential",
    "eslint:recommended",
    "@vue/eslint-config-prettier"
  ] // 💡 Replace `"extends":·[⏎····"plugin:vue/vue3-essential",⏎····"eslint:recommended",⏎····"@vue/eslint-config-prettier"⏎··` with `extends:·['plugin:vue/vue3-essential',·'eslint:recommended',·'@vue/eslint-config-prettier'`
}
```

I think that's caused by `JSON.stringify()` in the following line:

https://github.com/vuejs/create-vue/blob/58c4fdb5c53a5019401a86568868ce9cf255a3ce/utils/renderEslint.ts#L65

### Steps to reproduce

1. Scaffold a project with `create-vue`:

    ```sh
    npm init vue my-project # select ESLint and Prettier options
    ```

2. Open the project in VS Code.

3. In VS Code, install the official [ESLint extension](https://marketplace.visualstudio.com/items?itemName=dbaeumer.vscode-eslint) (if not already).

5. Observe the linter warnings in all `.vue` files, and in `.eslintrc.cjs`.

### Workaround

Copy `create-vue`'s [`.prettierrc`](https://github.com/vuejs/create-vue/blob/afddba9/.prettierrc) (which I presume was used to format the templates here) to the scaffolded project's root directory, but I think that should be done by `create-vue` automatically.

tony19:
Changes:

 * Scaffolds `.prettierrc` if Prettier is selected at prompt
 * Removes quotes from config object's keys in scaffolded `.eslintrc.cjs`

fix #116

### DIFF
--- a/index.ts
+++ b/index.ts
@@ -314,6 +314,9 @@ async function init() {
   if (needsEslint) {
     renderEslint(root, { needsTypeScript, needsCypress, needsCypressCT, needsPrettier })
   }
+  if (needsPrettier) {
+    render('config/prettier')
+  }
 
   // Render code template.
   // prettier-ignore
--- a/template/config/prettier/.prettierrc
+++ b/template/config/prettier/.prettierrc
@@ -1,0 +1,7 @@
+{
+  "semi": false,
+  "tabWidth": 2,
+  "singleQuote": true,
+  "printWidth": 100,
+  "trailingComma": "none"
+}
--- a/utils/renderEslint.ts
+++ b/utils/renderEslint.ts
@@ -62,7 +62,12 @@ function configureEslint({ language, styleGuide, needsPrettier, needsCypress, ne
     addEslintDependency('@rushstack/eslint-patch')
     configuration += `require("@rushstack/eslint-patch/modern-module-resolution");\n\n`
   }
-  configuration += `module.exports = ${JSON.stringify(config, undefined, 2)}\n`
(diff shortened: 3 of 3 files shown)


cexbrayat:
As explained in the linked issue, I think this is already solved in v3.8.0, so I'll close

PR #120, opened 2022-07-07
fix: modify the condition of log `lint`

liruifengv:
I think, it is more reasonable to change the condition of log `lint` from `needsPrettier ` to `needsEslint `


sodatea:
The default ESLint config doesn't include any style preferences, thus no need to run lint after creation.

PR #121, opened 2022-07-11
perf: exclude transpiled prompts code

sapphi-red:
Since this package supports Node 14+, the transpiled `prompts` code is not used.
https://github.com/terkelg/prompts/blob/a0c1777ae86d04e46cb42eb3af69ca74ae5d79e2/index.js#L11-L14
This PR removes that from bundle.

|                        | before     | after         |         diff |
|----------------|-----------:|-----------:|----------:|
|gz                    | 53,746B   | 39,690B   | -14,056B (-26.2%) |
|node_modules| 319,900B | 232,067B | -87,833B (-27.5%) |

### DIFF
--- a/scripts/build.mjs
+++ b/scripts/build.mjs
@@ -33,6 +33,16 @@ await esbuild.build({
   target: 'node14',
 
   plugins: [
+    {
+      name: 'alias',
+      setup({ onResolve, resolve }) {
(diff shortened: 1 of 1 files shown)


sodatea (approved):
Cool! Never thought about that!

PR #138, opened 2022-08-09
fix: the .git should be ignored when emptying the dir

btea:


sodatea (approved):
Thanks! This makes sense

Issue #82, opened 2022-03-26
Prompt user to select an ESLint based lint/format config (eg Airbnb) like Vue CLI


lloydjatkinson:
Hi,

Thank you for creating this tool!

It would be really great if `create-vue` could setup and configure (including with JS *or* TS support) a chosen ruleset - like Vue CLI does.

For example when using Vue CLI a choice is presented:

![image](https://user-images.githubusercontent.com/5633938/160260799-c5a1abe6-2a4b-4df9-ab5d-0ab914c0f78e.png)

`vue-create` today:

![image](https://user-images.githubusercontent.com/5633938/160260859-c0dd0103-7795-4acd-af12-88a9897d1e02.png)

I have looked into how this can be setup and honestly it's pretty hard to understand - ESLint is powerful but hard to configure **correctly**. There appears to be many possible solutions but once Typescript is in the mix the complexity increases for anyone, like me, that isn't that familiar with the really low level details of configuring ESLint. Even if someone is able to work through the (many) conflicting articles on setting this up there is still the reality of needing to go through the same complexity for every new project.

One of the great features of Vue CLI is that this is already configured for a user if they pick TS, ESLint, Aibnb.

Please could this prompt be added to `create-vue`? (and the `Standard` standard too for users of that)

Reidond:
it's not that hard to bootstrap eslint config on your own, you could even make your own packages with predefined config.
i personally use eslint + prettier combo and it works great

Reidond:
### Packages
```json
{
    "eslint": "^8.10.0",
    "eslint-config-prettier": "^8.4.0",
    "eslint-plugin-prettier": "^4.0.0",
    "eslint-plugin-vue": "^8.5.0",
    "prettier": "2.5.1",
    "vue-eslint-parser": "^8.3.0"
}
```
### Configs
.eslintrc.js
```js
module.exports = {
  root: true,
  env: {
    browser: true,
    node: true,
    es6: true,
    "vue/setup-compiler-macros": true,
  },
  parser: "vue-eslint-parser",
  parserOptions: {
    sourceType: "module",
    ecmaVersion: "latest",
  },
  extends: [
    "eslint:recommended",
    "plugin:vue/vue3-recommended",
    "plugin:prettier/recommended",
  ],
  rules: {
    "vue/multi-word-component-names": "off",
  },
};
```
.prettierrc.json
```json
{
  "trailingComma": "all",
  "tabWidth": 2,
  "semi": true,
[...]

PR #147, opened 2022-08-25
feat(eslint): support choose a eslint config

liruifengv:


liruifengv:
Seems to be a duplicate of https://github.com/vuejs/create-vue/issues/82

PR #148, opened 2022-08-26
Default open page.

zhangenming:
I think it's a useful feature...

### DIFF
--- a/template/base/vite.config.js
+++ b/template/base/vite.config.js
@@ -10,5 +10,8 @@ export default defineConfig({
     alias: {
       '@': fileURLToPath(new URL('./src', import.meta.url))
     }
+  },
+  server: {
+    open: true
   }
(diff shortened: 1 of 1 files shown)


linghengqian:
I don't think so.  At least this is false in webpack devserver, which can cause confusion for newbies migrating to vite.

zhangenming:
how about add it, but let it be comment. serve as a reminder
For me, I didn't even know about it until a long time ago

linghengqian:
> how about add it, but let it be comment. serve as a reminder
> For me, I didn't even know about it until a long time ago

I just thought, this is actually what vite's documentation mentions.  This is not a reasonable tip.

btea:
Vite already supports the shortcut key `o` to open in the browser. Therefore, I think this is unnecessary.

PR #151, opened 2022-08-31
tsconfig.vite.ts should include only the test files

segevfiner:
Otherwise node and jsdom types are available in ts and vue files. Whoops.

### DIFF
--- a/template/tsconfig/vitest/tsconfig.vitest.json
+++ b/template/tsconfig/vitest/tsconfig.vitest.json
@@ -1,5 +1,6 @@
 {
   "extends": "./tsconfig.app.json",
+  "include": ["src/**/__tests__/*"],
   "exclude": [],
   "compilerOptions": {
     "composite": true,


segevfiner:
Oh wait. This will cause the default type-check command to only check the types of the test files. We will have to modify that as well.

segevfiner:
Argh. There is a bit of a problem here. Type checking with just using `tsconfig.vite.json` won't work as the test files include the app's code, & `tsc/vue-tsc --build --noEmit` is not supported, as that mode relies on .d.ts output files to compile each project separately and type check them agains't each other. This might mean we will need to not use `noEmit` and use `declarationDir`/`outDir`/`emitDeclarationOnly` to some dedicated directory to run type checking.

Opinions anyone. @sodatea @cexbrayat

segevfiner:
I tried swapping the order of `tsconfig.app.json` & `tsconfig.vitest.json` in `tsconfig.json` and VS Code decided to use `tsconfig.app.json` instead. 🤷‍♂️  I still wonder if that's a proper fix or if there is a way to do this properly. e.g. Some variant of `tsc --build` that does work for only type checking maybe by emitting temporary .d.ts file to some .gitignored folder.

segevfiner:
Oh great. The resolution order is flipped between Volar and the builtin tsserver. Volar will prefer the first matching tsconfig, and tsserver the last!

PR #154, opened 2022-09-05
fix: remove console.log

cexbrayat:
The ESLint config is currently logged out when a new project is generated.


sodatea:
Holy ****… What a dumb mistake I've made🤦‍♂️

Background:
I made the code changes more than 1 week earlier and then I went to do something else.
Then yesterday I decided to ship a patch, but I chose to release it from my local branch, thinking these changes should've already been tested well…
Then all these dumb bugs were introduced.

Guess I still have a lot to improve on my workflow…

PR #161, opened 2022-09-11
fix: the projectName should not be `.` in README if created in current directory

btea:


sodatea:
There's no such bug. Because when the user passes `.` to the command line, the `projectName` prompt will be skipped; therefore `result.projectName` will be `undefined`.

btea:
The `projectName` prompt doesn't seem to be skipped when the user enters `.`, so the value of `projectName` is `.`. Therefore, the title in README is `.`.
image

sodatea:
I can't reproduce it

btea:
Follow the steps shown in the figure to reproduce
image

Maybe it's related to the system environment, I'm using windows 🤔

sodatea:
😂 You explicitly used `.` as the project **name**…
It's a name, not a path. I don't expect people to use it this way.

The way to create the project in the current directory should be `pnpm create vue@3 .`

btea:
🤣 Embarrassing, I always use this to create projects in the current folder.

PR #163, opened 2022-09-15
chore: format code & remove extra blank lines

btea:


template/config/cypress-ct/cypress/support/component-index.html
```diff
@@ -1,9 +1,9 @@
 <!DOCTYPE html>
 <html>
   <head>
-    <meta charset="utf-8">
-    <meta http-equiv="X-UA-Compatible" content="IE=edge">
-    <meta name="viewport" content="width=device-width,initial-scale=1.0">
+    <meta charset="utf-8" />
```
sodatea, line 4: `<meta>` is a [void element](https://developer.mozilla.org/en-US/docs/Glossary/Void_element), thus there should be no end tag.
https://html.spec.whatwg.org/multipage/semantics.html#the-meta-element

sodatea (changes requested):

template/config/cypress-ct/cypress/support/component-index.html
btea: OK, thanks for the documentation instructions. The end tag here is added after running `pnpm format`, maybe this is a problem with `prettier`? 🤔

sodatea: Yeah, tracked at https://github.com/prettier/prettier/issues/5246
In that case, I'd like to add `**/*.html` to the project `.prettierignore` for now.

sodatea (approved):

PR #166, opened 2022-09-24
Update base.css

TrumanGao:


sodatea (commented):
Could you elaborate why this change is necessary?

TrumanGao:
> Could you elaborate why this change is necessary?

In some third-party dependencies (e.g. [monaco-editor](https://github.com/microsoft/monaco-editor)), the offsets of some positioned elements are not relative to their parent nodes but their ancestor nodes, cuz their parent nodes have no position attribute. Therefor, the original code may lead to certain unexpected offsets.

sodatea (changes requested):
Makes sense.
But this change breaks the styles of the landing page. We need to fix that before removing the global reset.

PR #181, opened 2022-10-19
feat!: make `type-check` command more comprehensive

### LINKED ISSUE #159: VS Code builtin tsserver/TypeScript support picks tsconfig.vitest.json instead of tsconfig.app.json for ts files
Follow up to talking to myself in https://github.com/vuejs/create-vue/pull/151 😛 

Create a new create-vue project with everything but tsx selected. In this configuration, VS Code's builtin tsserver/TypeScript support seems to pick `tsconfig.vitest.json`, instead of `tsconfig.app.json`. It seems the logic to pick them is different between Volar and tsserver, as changing their order in `tsconfig.json` causes Volar to be the one to select `tsconfig.vitest.json` instead of `tsconfig.app.json`.

Adding the following to `tsconfig.vitest.json`:
```json
  "references": [
    { "path": "./tsconfig.app.json" }
  ]
```

Seems to make VS Code select `tsconfig.app.json` for ts files properly, but this is some pure guess work and I'm not sure where or how the logic to select an appropriate tsconfig works.

In addition:
1. We are using `vue-tsc --noEmit -p tsconfig.vitest.json --composite false` to type check, but in case the user wants to add additional projects to the project references config, this won't type check them. Apparently `--build`, `composite`, and project references don't support `noEmit`, so we need to use `emitDeclarationOnly`, `declarationDir` instead to make it emit types to some folder solely for the purpose of type checking. On the plus side, in this mode it reuses them when it can for faster type checking (It also enables `incremental`).
2. `tsconfig.vitest.json` has `"lib": []`, but this should probably be `"lib": ["ES2016"]` to match `tsconfig.app.json` just without DOM.

So basically the following changes to the generated project seems to fix those things for me:
```diff
diff --git a/.gitignore b/.gitignore
index 38adffa..b2e683b 100644
--- a/.gitignore
+++ b/.gitignore
@@ -13,6 +13,8 @@ dist
 dist-ssr
 coverage
 *.local
+types
+*.tsbuildinfo
 
 /cypress/videos/
 /cypress/screenshots/
diff --git a/package.json b/package.json
index 4c8c275..1a7cbc7 100644
--- a/package.json
+++ b/package.json
@@ -9,7 +9,7 @@
     "test:e2e": "start-server-and-test preview http://localhost:4173/ 'cypress open --e2e'",
     "test:e2e:ci": "start-server-and-test preview http://localhost:4173/ 'cypress run --e2e'",
     "build-only": "vite build",
-    "type-check": "vue-tsc --noEmit -p tsconfig.vitest.json --composite false",
+    "type-check": "vue-tsc --build",
     "lint": "eslint . --ext .vue,.js,.jsx,.cjs,.mjs,.ts,.tsx,.cts,.mts --fix --ignore-path .gitignore"
   },
   "dependencies": {
diff --git a/tsconfig.app.json b/tsconfig.app.json
index cdbea1d..0e1bcdc 100644
--- a/tsconfig.app.json
+++ b/tsconfig.app.json
@@ -3,6 +3,8 @@
   "include": ["env.d.ts", "src/**/*", "src/**/*.vue"],
   "exclude": ["src/**/__tests__/*"],
   "compilerOptions": {
+    "emitDeclarationOnly": true,
+    "declarationDir": "types",
     "composite": true,
     "baseUrl": ".",
     "paths": {
diff --git a/tsconfig.config.json b/tsconfig.config.json
index c2d3a30..8577b3c 100644
--- a/tsconfig.config.json
+++ b/tsconfig.config.json
@@ -2,6 +2,8 @@
   "extends": "@vue/tsconfig/tsconfig.node.json",
   "include": ["vite.config.*", "vitest.config.*", "cypress.config.*"],
   "compilerOptions": {
+    "emitDeclarationOnly": true,
+    "declarationDir": "types",
     "composite": true,
     "types": ["node"]
   }
diff --git a/tsconfig.vitest.json b/tsconfig.vitest.json
index d080d61..d0d8b87 100644
--- a/tsconfig.vitest.json
+++ b/tsconfig.vitest.json
@@ -2,8 +2,13 @@
   "extends": "./tsconfig.app.json",
   "exclude": [],
   "compilerOptions": {
+    "emitDeclarationOnly": true,
+    "declarationDir": "types",
     "composite": true,
-    "lib": [],
+    "lib": ["ES2016"],
     "types": ["node", "jsdom"]
-  }
+  },
+  "references": [
+    { "path": "./tsconfig.app.json" }
+  ]
 }
```

@sodatea

sodatea:
Fixes #159

### DIFF
--- a/template/tsconfig/cypress-ct/tsconfig.cypress-ct.json
+++ b/template/tsconfig/cypress-ct/tsconfig.cypress-ct.json
@@ -2,8 +2,8 @@
   "extends": "./tsconfig.app.json",
   "include": [
     "env.d.ts",
-    "src/**/*",
-    "src/**/*.vue",
+    "src/**/__tests__/**/*",
+    "src/**/__tests__/**/*.vue",
     "cypress/support/component.*",
     "cypress/support/commands.ts"
   ],
--- a/template/config/typescript/package.json
+++ b/template/config/typescript/package.json
@@ -2,7 +2,7 @@
   "scripts": {
     "build": "run-p type-check build-only",
     "build-only": "vite build",
-    "type-check": "vue-tsc --noEmit"
+    "type-check": "run-p type-check:*"
   },
   "devDependencies": {
     "@types/node": "^16.11.65",
--- a/template/tsconfig/base/package.json
+++ b/template/tsconfig/base/package.json
@@ -1,4 +1,8 @@
 {
+  "scripts": {
+    "type-check:app": "vue-tsc --noEmit",
+    "type-check:config": "vue-tsc --noEmit -p tsconfig.config.json --composite false"
+  },
   "devDependencies": {
     "@vue/tsconfig": "^0.1.3"
   }
--- a/template/tsconfig/cypress-ct/package.json
+++ b/template/tsconfig/cypress-ct/package.json
@@ -1,5 +1,7 @@
 {
   "scripts": {
-    "type-check": "vue-tsc --noEmit -p tsconfig.cypress-ct.json --composite false"
+    "type-check": "run-p type-check:*",
+    "type-check:app": "vue-tsc --noEmit -p tsconfig.app.json --composite false",
+    "type-check:cypress-ct": "vue-tsc --noEmit -p tsconfig.cypress-ct.json --composite false"
   }
 }
--- a/template/tsconfig/vitest/package.json
+++ b/template/tsconfig/vitest/package.json
@@ -1,6 +1,8 @@
 {
   "scripts": {
-    "type-check": "vue-tsc --noEmit -p tsconfig.vitest.json --composite false"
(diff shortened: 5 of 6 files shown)


template/tsconfig/cypress-ct/tsconfig.cypress-ct.json
```diff
@@ -2,8 +2,8 @@
   "extends": "./tsconfig.app.json",
   "include": [
     "env.d.ts",
-    "src/**/*",
-    "src/**/*.vue",
+    "src/**/__tests__/**/*",
```
sodatea, line 5: By only including those under `__tests__` by default, we avoid type-checking most of the component files twice; only the components that are also `import`ed by Cypress Component Testing specs would be type-checked again in the `type-check:cypress-ct` task.

PR #182, opened 2022-10-19
refactor!: use prod bundle & preview server for playwright testing on CI

sodatea:
To make it consistent with the Cypress configurations.

Background: https://twitter.com/haoqunjiang/status/1582692547188920322

Alternative: use `npm run dev` in Cypress configuration, as suggested in https://github.com/vuejs/create-vue/pull/39

### DIFF
--- a/template/config/playwright/playwright.config.ts
+++ b/template/config/playwright/playwright.config.ts
@@ -102,7 +102,7 @@ const config: PlaywrightTestConfig = {
      * Use the dev server by default for faster feedback loop.
      * Use the preview server on CI for more realistic testing.
      */
-    command: process.env.CI ? 'npm run preview --port 5173' : 'npm run dev',
+    command: process.env.CI ? 'npm run preview -- --port 5173' : 'npm run dev',
     port: 5173,
     reuseExistingServer: !process.env.CI
   }
--- a/template/config/playwright/playwright.config.ts
+++ b/template/config/playwright/playwright.config.ts
@@ -102,7 +102,7 @@ const config: PlaywrightTestConfig = {
      * Use the dev server by default for faster feedback loop.
      * Use the preview server on CI for more realistic testing.
      */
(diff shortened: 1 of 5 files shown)


sodatea:
Cc @mxschmitt @lmiller1990
Do you have any preferences on this configuration?

mxschmitt (commented):
sounds good to me, but it requires an additional step for running e2e tests.

I feel like customers would run into issues, e.g. that they run tests against outdated build directory, or are not easily able to run tests against a dev server (not easily means they would need to change playwright config file baseURL and webServer run command).

Was also thinking about something like this: only run the webserver when it's on CI, for the other cases rely on the user to use npm run dev / npm run build+preview

sodatea:
> Was also thinking about something like this: only run the webserver when it's on CI, for the other cases rely on the user to use npm run dev / npm run build+preview

I feel this is the better option. Updated the PR accordingly.
And I also opened a PR to apply the similar changes to the Cypress templates: https://github.com/vuejs/create-vue/pull/183

sodatea:
> Was also thinking about something like this: only run the webserver when it's on CI, for the other cases rely on the user to use npm run dev / npm run build+preview

I feel this is a better option.
But to make the configuration simpler, I think we can still reuse the `CI` environment variable to switch between different servers, they can share the same port though.
I've updated the PR accordingly. What do you think?

template/config/playwright/playwright.config.ts
```diff
@@ -98,7 +98,11 @@ const config: PlaywrightTestConfig = {
 
   /* Run your local dev server before starting the tests */
   webServer: {
-    command: 'npm run dev',
+    /**
+     * Use the dev server by default for faster feedback loop.
+     * Use the preview server on CI for more realistic testing.
+     */
```
mxschmitt, line 104:
```suggestion
    Playwright will re-use the local server if there is already a dev-server running.
     */
```

something like that is missing, then its excellent!

mxschmitt (approved):

PR #183, opened 2022-10-20
refactor!: run cypress headlessly by default; add test:*:dev commands 

sodatea:
for headed environment and testing against dev server.

I think this better reflects the real-world use case.

People don't usually need an app window to run the tests against
the production bundle, but faster feedback loops with dev server are
useful.

Inspired by @elevatebart 's comment on twitter: https://twitter.com/elevatebart/status/1582790882088009728

### DIFF
--- a/template/config/cypress/package.json
+++ b/template/config/cypress/package.json
@@ -1,7 +1,7 @@
 {
   "scripts": {
-    "test:e2e": "start-server-and-test preview http://localhost:4173/ 'cypress open --e2e'",
-    "test:e2e:ci": "start-server-and-test preview http://localhost:4173/ 'cypress run --e2e'"
+    "test:e2e": "start-server-and-test preview :4173 'cypress run --e2e'",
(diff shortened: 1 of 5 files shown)


template/config/cypress/package.json
```diff
@@ -1,7 +1,7 @@
 {
   "scripts": {
-    "test:e2e": "start-server-and-test preview http://localhost:4173/ 'cypress open --e2e'",
-    "test:e2e:ci": "start-server-and-test preview http://localhost:4173/ 'cypress run --e2e'"
+    "test:e2e": "start-server-and-test preview :4173 'cypress run --e2e'",
+    "test:e2e:dev": "start-server-and-test 'vite dev --port 4173' :4173 'cypress open --e2e'"
```
sodatea, line 4: I chose to force the dev server to run on 4173 instead of 5173 as in the playwright PR https://github.com/vuejs/create-vue/pull/182 because `start-server-and-test` doesn't have an option like `reuseExistingServer`, therefore has a higher chance of port collisions during dev.

PR #202, opened 2022-11-30
fix: stringify targetDir if present

### LINKED ISSUE #201: `npm create vue@3 1` throws`ERR_INVALID_ARG_TYPE`
`npm create vue@3 1` / `yarn create vue 1` / `pnpm create vue 1` throws the following error.

```
TypeError [ERR_INVALID_ARG_TYPE]: The "path" argument must be of type string. Received type number (1)
    at new NodeError (node:internal/errors:387:5)
    at validateString (node:internal/validators:121:11)
    at Object.join (node:path:429:7)
    at init (C:\Users\green\AppData\Local\npm-cache\_npx\d8a5db81a46afa44\node_modules\create-vue\outfile.cjs:4041:22)
    at processTicksAndRejections (node:internal/process/task_queues:96:5) {
  code: 'ERR_INVALID_ARG_TYPE'
}
```

Original issue: https://github.com/vitejs/vite/issues/11129

luo3house:
Fix #201 

The targetDir retrieved from `minimist` may be an number, and nodejs `path.join` does not accept number as argument.

### DIFF
--- a/index.ts
+++ b/index.ts
@@ -98,6 +98,9 @@ async function init() {
     ) === 'boolean'
 
   let targetDir = argv._[0]
+  if (targetDir !== undefined) {
+    targetDir = String(argv._[0])
+  }
   const defaultProjectName = !targetDir ? 'vue-project' : targetDir
 
   const forceOverwrite = argv.force


luo3house:
It seems the code is a bit ugly because package `minimist` provides a `string` option for argument stringifying.

PR #203, opened 2022-12-02
fix: project name with only numbers as an argument (#201)

luo3house:
Supersedes #202

Use a `string` option of `minimist` to stringify the potential numeric project name.


sodatea (approved):
Thanks!

PR #210, opened 2022-12-10
Remove unused namespace declaration

MartijnCuppens:


sodatea (approved):
Thanks!
It's somehow ironic that a compressor would insert a redundant namespace on its own 😂

PR #225, opened 2023-02-01
Add padding to a tag to make more visualised

zobay:
## Description
- Add padding to anchor tag to make more visible.

## Before: (Without padding)
Screenshot 2023-02-01 at 12 46 39 PM

## After: (With padding)
Screenshot 2023-02-01 at 12 46 22 PM


cexbrayat (approved):
LGTM, thanks @zobay

PR #227, opened 2023-02-05
fix: use a standalone config file for the vitest

### LINKED ISSUE #221: vitest - results.json gets created in ./src/node_modules
## Problem:
I created today a new project with create-vue and when directly afterwards run `test:unit` it creates under "./src" a folder "node_modules/.vitest" and puts the results.json there.  
Also the coverage folder is created below ."/src" if i run tests with activated coverage.

I would expect that the files created by running unit tests are directly at the project root folder "./".

## Problem reproduction
For reproduction:
I am working in a Windows WSL environment (Debian) and am using node 16.13.2.
I called `npm init vue@latest`
Then choose to use
[X] TypeSCript
[ ]No JSX
[X] Router
[X] Pinia
[X] Vitest
[X] E2E = Cypress
[X] ESLINT
[X] Prettier

I switched into the created project folder, run `npm install` and then `npm run unit:test`.
And then the folder/file gets created under "./src".
Also the coverage folder is created below ."/src" if i run tests with activated coverage.

## Possible Reason:
In the package.json the script is
"test:unit": "vitest --environment jsdom --root src/",``
Which makes "./src" the root.

## Eventual solution
Instead change the options to:
"test:unit": "vitest --environment jsdom --root ./ --dir ./src",``

Or perhaps directly move the root setting directly into the vite.config.ts file.
test: {
  root: fileURLToPath(new URL('./', import.meta.url)),
},

btea:
…eport file

fix #221


btea:
@sodatea Please review this pr, thanks.

sodatea (approved):
Thanks for the help and sorry for the late review.

The standalone configuration is a bit more verbose than I would like. But so far I haven't found a better solution.

This also seems to be the most scalable way to configure Vitest. So I can settle for it.

PR #237, opened 2023-03-02
fix: change link to cypress api docs.

ByMykel:
The current link redirects to a 404 page.

image

I've change the link to [The API Docs](https://docs.cypress.io/api/commands/and)


sodatea (approved):
Thanks for spotting this!
I have updated the URL to use the `on` link, which seems more permanent and widely used in their repositories.

PR #249, opened 2023-04-07
chore: playwright component testing description

sand4rt:


sodatea:
Thanks for the PR!
But IMO, if we are to tell the users that component testing is also supported, we need to add a template for it first, otherwise users might be confused when they see no reference to the component testing after scaffolding the project.

PR #250, opened 2023-04-11
feat: if use playwright, add related extension recommendations

### LINKED ISSUE #248: Add `"ms-playwright.playwright"` in `.vscode/extensions.json` when Playwright is chosen
I thought it would be good to have `ms-playwright.playwright` in `.vscode/extensions.json` by default when I choose Playwright as e2e solution.

# What I get

1. `npm init vue@latest`
2. `Add an End-to-End Testing Solution? › Playwright`
3. `.vscode/extensions.json` not containing `"ms-playwright.playwright"`

```json
{
  "recommendations": [
    "Vue.volar",
    "Vue.vscode-typescript-vue-plugin"
  ]
}
```

# What I want

1. `npm init vue@latest`
2. `Add an End-to-End Testing Solution? › Playwright`
3. `"ms-playwright.playwright"` in `.vscode/extensions.json`

```json
{
  "recommendations": [
    "Vue.volar",
    "Vue.vscode-typescript-vue-plugin",
    "ms-playwright.playwright"
  ]
}
```

btea:
close #248

### DIFF
--- a/index.ts
+++ b/index.ts
@@ -307,13 +307,6 @@ async function init() {
     const templateDir = path.resolve(templateRoot, templateName)
     renderTemplate(templateDir, root)
   }
-  const addPlaywrightExtensions = function addPlaywrightExtensions() {
-    const extensionDir = path.resolve(root, '.vscode/extensions.json')
-    const json = fs.readFileSync(extensionDir, 'utf-8')
-    const obj = JSON.parse(json)
-    obj.recommendations.push('ms-playwright.playwright')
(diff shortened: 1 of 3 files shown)


sodatea (changes requested):
Thanks! I like the idea of adding more extension recommendations. But this approach doesn't look scalable to me.
How about putting `.vscode/extensions.json` in the playwright template folder, and adding a special merging strategy for this file (just like what we already did for `package.json`). Then we can easily add more extension recommendations (for example, the vitest extension) without adding too much logic code.

sodatea:
No worries, it's just that I messed up the Cypress caches in CI.

PR #274, opened 2023-05-12
fix: use `--build` for type-checking to be exaustive and less-fragile

### LINKED ISSUE #267: Npm script `type-check` is not working
Steps to reproduce:
* Create empty directory and navigate to it
* Run `npm init`
* Install `create-vue@latest`, currently it is `3.6.3`
* Run: `npx create-vue --ts --router --pinia --eslint-with-prettier source` and switch to `source` directory
*  Make some errors in files
* Run `npm run type-check`

There are no errors reported by command.

sodatea:
Fixes #267

Thanks @segevfiner for noticing this new feature!

Currently there's still a small annoyance that `vue-tsc` would output an extra error message when there're type errors in `.vue` files https://github.com/vuejs/language-tools/issues/2622 But it works well if there's no error.

I've already submitted a PR to address it https://github.com/vuejs/language-tools/pull/3176

### DIFF
--- a/template/base/_gitignore
+++ b/template/base/_gitignore
@@ -26,3 +26,5 @@ coverage
 *.njsproj
 *.sln
 *.sw?
+
+*.tsbuildinfo
--- a/template/config/typescript/package.json
+++ b/template/config/typescript/package.json
@@ -2,7 +2,7 @@
   "scripts": {
     "build": "run-p type-check build-only",
     "build-only": "vite build",
-    "type-check": "vue-tsc --noEmit"
+    "type-check": "vue-tsc --build"
   },
   "devDependencies": {
     "@types/node": "^18.16.3",
--- a/template/tsconfig/base/tsconfig.app.json
+++ b/template/tsconfig/base/tsconfig.app.json
@@ -4,6 +4,7 @@
   "exclude": ["src/**/__tests__/*"],
   "compilerOptions": {
     "composite": true,
+    "noEmit": true,
     "baseUrl": ".",
     "paths": {
       "@/*": ["./src/*"]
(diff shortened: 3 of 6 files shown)


segevfiner:
Actually, they might have changed things so that it does run but doesn't type check packages with `noEmit` at a later releas...

Be sure we actually get errors with this. I have requested https://github.com/microsoft/TypeScript/issues/53979 for them to make this work.

sodatea:
What do you mean by "might have changed things"?

I tested this project https://github.com/sodatea/no-emit-test with both 5.0.4 and 5.1.0-dev.20230512, both reporting the same correct number of errors - though in subsequent calls the error location is lost for `./src/App.vue`.
That can be fixed by removing `*.tsbuildinfo` each time before type-checking?

sodatea:
> That can be fixed by removing `*.tsbuildinfo` each time before type-checking?

`vue-tsc --build --force` seems to work quite reliably.

segevfiner:
I tried configuring this myself once before but got no errors, but maybe I did something wrong or it was a different issue.

segevfiner:
It might have been the `--force` thing

sodatea:
@cexbrayat This change could help catch the issue in https://github.com/vuejs/create-vue/pull/386
But I'm unsure if it's worth bringing the extra complexity to average users.

cexbrayat (approved):
I gave it a try, and I think it works well 👍

One note though: all the errors are logged twice. But I'm not sure if this is really an issue.

cexbrayat (changes requested):
Looks like `typescript-jsx-nightwatch` is failing when running `pnpm test` locally. There is probably a `noEmit` to add in another file

cexbrayat (approved):
👍

bmulholland:
FYI: I think this setting is incompatible with any JS in the project, though. I have some legacy JS files in my app, so I have allowJS: true. If I try running type checking, I get "js emit is not supported." And if I add `--noEmit` again, "Compiler option '--noEmit' may not be used with '--build'"

PR #281, opened 2023-05-15
fix: `WelcomeItem` component style

yuyinws:
Due to the lack of the CSS property `position: relative` in these files, the initialization template icon cannot be displayed correctly.

image


yuyinws:
repeat to #278

PR #289, opened 2023-05-30
fix: relative position in `i` elements in WelcomeItem

dennybiasiolli:
Before:
image

After:
image


btea:
Looks like it's been fixed by #278.

cexbrayat:
Yes, I think we can close this one. Thanks @dennybiasiolli  !

PR #310, opened 2023-07-07
docs(readme): update install command

blowsie:
Updates install command to be inline with Vue Certification questions and avoids using npm alias


sodatea:
Actually, it's an intentional choice.
Because `yarn` and `pnpm` only support the `create` subcommand.
`init` has different meanings for them.

So using `create` leads to the least confusion.

As for the Vue Certification questions, I'll talk to them to keep things consistent.

PR #338, opened 2023-09-07
fix(config/typescript): pass build arguments to build-only command

wsehl:
I was configuring builds with different modes and couldn't run ```yarn build --mode preview```  as I would do normally with simple ```vite build``` in my "build" script because of ```ERROR: Invalid Option: --mode```

Had to spend some time figuring this out, guess it may be a helpfull addition that will prevent someone from getting a similar issue


sodatea (approved):
Thanks! This is indeed very helpful!

PR #348, opened 2023-09-25
feat: i18n solution for prompts messages

luventa:
Straightforward i18n solution for prompts messages

### DIFF
--- a/index.ts
+++ b/index.ts
@@ -15,6 +15,7 @@ import renderTemplate from './utils/renderTemplate'
 import { postOrderDirectoryTraverse, preOrderDirectoryTraverse } from './utils/directoryTraverse'
 import generateReadme from './utils/generateReadme'
 import getCommand from './utils/getCommand'
+import getLanguage from './utils/getLanguage'
 import renderEslint from './utils/renderEslint'
 import { FILES_TO_FILTER } from './utils/filterList'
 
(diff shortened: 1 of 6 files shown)


luventa:
ci test is blocked because of this:

sodatea (changes requested):
Thanks! I've been thinking about this feature for a while.
However, I don't think it should be an additional option. Many people don't wanna bother selecting one at all.

Please make it implicit by reading the user's language setting from the environment variables.

luventa:
> Thanks! I've been thinking about this feature for a while. However, I don't think it should be an additional option. Many people don't wanna bother selecting one at all.
> 
> Please make it implicit by reading the user's language setting from the environment variables.

No problem, already updated. Is there anything that needs improvement?

index.ts
```diff
@@ -148,110 +151,118 @@ async function init() {
...
         },
         {
           name: 'needsTypeScript',
           type: () => (isFeatureFlagsUsed ? null : 'toggle'),
-          message: 'Add TypeScript?',
+          message: language.needsTypeScript.message,
           initial: false,
-          active: 'Yes',
-          inactive: 'No'
+          active: language.needsTypeScript.toggleOptions[0],
```
sodatea, line 194: This is hard to understand at a glance.
It needs to be more semantic, such as `toggleOptions['active']`. For now, all the `toggleOptions` are the same, so a global one should suffice, i.e. `language.defaultToggleOptions['active']`

index.ts
```diff
@@ -148,110 +151,118 @@ async function init() {
...
-              title: 'Cypress',
+              title: language.needsE2eTesting.selectOptions[0].title,
+              value: false
+            },
+            {
+              title: language.needsE2eTesting.selectOptions[1].title,
               description: answers.needsVitest
                 ? undefined
-                : 'also supports unit testing with Cypress Component Testing',
+                : language.needsE2eTesting.selectOptions[1].desc,
```
sodatea, line 243: Again, array indexes are hard to understand and maintain (when we need to sync modifications among different languages). Please use more semantic keys.

locales/zh-CN.json
```diff
@@ -0,0 +1,58 @@
+{
+  "projectName": {
+    "message": "请输入项目名称:"
```
sodatea, line 3: Please use the full-width colon character `：`
https://github.com/vuejs-translations/docs-zh-cn/wiki/%E7%BF%BB%E8%AF%91%E9%A1%BB%E7%9F%A5#22-%E6%A0%87%E7%82%B9%E7%AC%A6%E5%8F%B7

locales/zh-CN.json
```diff
@@ -0,0 +1,58 @@
...
+  },
+  "shouldOverwrite": {
+    "dirForPrompts": ["当前目录", "目标文件夹"],
+    "message": "非空，是否覆盖?",
```
sodatea, line 7: Please use the full-width question mark `？`

sodatea (changes requested):

sodatea (approved):
Thanks! Looks really good.

PR #357, opened 2023-10-20
feat: add invalid package name message translation

btea:


locales/fr-FR.json
```diff
@@ -10,7 +10,8 @@
     "message": "n'est pas vide. Supprimer les fichiers existants et continuer\u00a0?"
   },
   "packageName": {
-    "message": "Nom du package\u00a0:"
+    "message": "Nom du package\u00a0:",
+    "invalidMessage": "Invalide package.json nom du"
```
cexbrayat, line 14: In French, that would be `Le nom du package.json est invalide`

cexbrayat (changes requested):
The french translation should be updated otherwise looks good to me  👍

locales/fr-FR.json
btea: Thank you for your review, it has been revised.

cexbrayat (approved):

PR #369, opened 2023-11-08
feat: support lang option

btea:


cexbrayat (approved):

cexbrayat:
@sodatea @btea  I'm wondering something: is there documentation somewhere that lists the flags that can be used with create-vue?

I'm afraid that `--lang` will go unnoticed if we don't document it.

sodatea:
Why do we need this option? I think environment variable would suffice, if not more convenient (and with less maintenance burden).
```
LANG=zh_CN npm create vue@latest
```

sodatea:
> is there documentation somewhere that lists the flags that can be used with create-vue?

I didn't explicitly document them because I was initially concerned about the potential breaking changes. And later we indeed made one: `--with-tests` uses vitest now.
But as of now, I think we can document some of the more stable options.
I think `--with-tests` is still subject to change, and I haven't made my opinion on the `--eslint-with-prettier` feature. The rest should be stable enough now.

btea:
> Why do we need this option? I think environment variable would suffice, if not more convenient (and with less maintenance burden).

I think providing a `lang` option can make it easier for users to switch their location more autonomously and flexibly.
> ```
> LANG=zh_CN npm create vue@latest
> ```
Powershell does not seem to support this writing method. If it is passed successfully using bash, `Intl.DateTimeFormat().resolvedOptions().locale` will also overwrite the passed value.

CoolPlayLin:
>Powershell does not seem to support this writing method. If it is passed successfully using bash, Intl.DateTimeFormat().resolvedOptions().locale will also overwrite the passed value.

@btea 
No, powershell actually support setup env. you set env in `$env` var
So, you can simply replace `LANG=zh_CN npm create vue@latest` with `$env:LANG = "en_US" & npm create vue@latest` to choose the locale you want to use. But currently, `Intl.DateTimeFormat().resolvedOptions().locale` has the highest priority and it caused choosing locale fail.

PR #372, opened 2023-11-10
feat: add hint message translation

btea:


locales/fr-FR.json
```diff
@@ -30,6 +30,7 @@
   },
   "needsE2eTesting": {
     "message": "Ajouter une solution de test de bout en bout (e2e)\u00a0?",
+    "hint": "Utilisez les touches fléchées. Retourner pour soumettre",
```
btea, line 33: Regarding the French translation, I simply translated it with the help of a translation tool. Coudl you help check and modify it? @cexbrayat

cexbrayat: I would say `- Utilisez les flèches et appuyez sur la touche Entrée pour valider`
(I also added the missing dash at the beginning)

cexbrayat (approved):

locales/fr-FR.json
btea: I've updated as you suggested.

cexbrayat (approved):

PR #365, opened 2023-11-04
feat: make locales reusable & add zh-Hant locale

CoolPlayLin:

PR #375, opened 2023-11-11
feat: Add zh-HK locale

CoolPlayLin:


btea (changes requested):
@CoolPlayLin  It seems that the content is basically the same as the `zh-TW` file, and may be reused directly. 🤔

CoolPlayLin:
merged into #365

PR #383, opened 2023-11-23
fix: snapshot generation

cexbrayat:
It looks like the addition of nightwatch broke the snapshot generation, as `nightwatch` was added to the deny list (correctly), but it means that the _only_ combinations that are filtered are the ones with the 3 e2e test libraries at the same time.

We also need to filter out the combination of Cypress with Playwright, Cypress with Nightwatch, and Playwright with Nightwatch.

See https://github.com/vuejs/create-vue-templates/ that currently includes cypress-playwright for example.

This commit updates the script to remove all generated combinations, and to generate only the proper ones.


cexbrayat:
@sodatea This is a first step to support Vite 5: currently ,the `pnpm test` script is failing because the playground includes projects that can't work.
Then we need to fix the `*-nightwatch` projects. They aren't tested on CI, and they don't work out-of-the-box. I'll see what I can do when this PR lands.

sodatea (approved):

sodatea:
> Then we need to fix the `*-nightwatch` projects. They aren't tested on CI, and they don't work out-of-the-box. I'll see what I can do when this PR lands.

GitHub Actions limits the maximum number of jobs a matrix can generate, so I had to leave out some of the tests.
Maybe we can squash some small jobs into one and parallelize the rest; hopefully the CI time won't increase too much.

cexbrayat:
Thanks for the review. I'll fix the local `test` script first to make sure Nightwatch projects are OK, then we can take a look at the CI workflow 👍

PR #387, opened 2023-11-25
feat: generate start and test npm scripts

### LINKED ISSUE #303: Use well-defined npm script names instead of custom
## Description

npm provides both `npm start` and `npm test`. Use these instead of `npm run dev` and `npm run test:unit`.

## Considerations

Maybe there are different layers of tests. So then keep `test:unit` + others, and add `test: "run-p test:unit ..."`

szgabsz91:
Resolves #303


cexbrayat (commented):
Thanks for the PR @szgabsz91 

I think you can do simpler than this, by just adding the script directly to the `template/base/package.json` for `start`. You can probably do the same for `test` with the `template/**/package.json` of each library, and that would avoid adding the `addNpmScript` function.

You can then generate the snapshots locally with `pnpm snapshot` and check if the result is OK.

szgabsz91:
Thanks for the quick review, @cexbrayat, I added a new commit with the modifications, keeping the old commit just for reference. (If the PR gets accepted in the future, a squash might be necessary.)

Just some comments:

* I eliminated the `addNpmScript` function, but without it, I could only use the same value for the `dev` and `test` scripts as their referenced scripts, because I do not know in the template which package manager the developer uses. With the previous solution, if you used npm, it generated `npm run dev` for instance, or `yarn dev` in case of yarn, etc.
* I also needed to modify the order of rendered templates. Now the E2E test related templates are rendered first so that they do not overwrite the `test` script if unit tests are also generated.
* I tried to test the modified CLI locally and I think the generated scripts work correctly as you suggested in the issue, but maybe using real references (like `npm run dev`, `npm run test:unit`) would be more elegant. However, I don't have an idea how to do so without using a custom function or implementing some kind of template parameters like using mustache or a similar lib in the render function.

I don't know if the current solution is OK with you, but if not, I'll try to modify it later as you suggest.

scripts/test.mjs
```diff
@@ -37,4 +37,9 @@ for (const projectName of projects) {
     console.log(`Running unit tests in ${projectName}`)
     await $`pnpm test:unit`
   }
+
+  if ('test:unit' in packageJSON.scripts || 'test:e2e' in packageJSON.scripts) {
+    console.log(`Running the test script in ${projectName}`)
+    await $`pnpm test`
+  }
```
cexbrayat, line 44: I think this can be removed as the script already tests the unit and e2e tests, and it already takes a _really long_ time to execute locally.

index.ts
```diff
@@ -348,24 +348,24 @@ async function init() {
...
   }
+  if (needsVitest) {
+    render('config/vitest')
+  }
+  if (needsCypressCT) {
+    render('config/cypress-ct')
+  }
+  if (needsNightwatchCT) {
+    render('config/nightwatch-ct')
+  }
```
cexbrayat, line 368: I don't mind switching the order but I'm afraid this might break something. But if the `pnpm test` script still work then it should be OK.

template/config/cypress-ct/package.json
```diff
@@ -1,7 +1,8 @@
 {
   "scripts": {
     "test:unit": "cypress run --component",
-    "test:unit:dev": "cypress open --component"
+    "test:unit:dev": "cypress open --component",
+    "test": "cypress run --component"
```
cexbrayat, line 5: maybe instead of repeating the command it could just be `"test": "npm run test:unit"`?
Same in every template

cexbrayat (changes requested):
I think I like this way better 👍

template/config/cypress-ct/package.json
szgabsz91: OK! I thought at first that using `npm run` if the developer generated the project using `yarn` is a bad idea, but if not, I modify the templates.

szgabsz91:
I did the rebase yesterday, if I see it correctly, the branch is up to date with the remote main. I also resolved the conflicts around the nightwatch package.json files.

cexbrayat (approved):
LGTM, thanks @szgabsz91 

I'll let @sodatea take a look and merge if he feels this is interesting

szgabsz91:
Thanks!

marcosdissotti (approved):

marcosdissotti (approved):

marcosdissotti (approved):

szgabsz91:
FYI: I just rebased the branch, now it's up-to-date again. :)

PR #388, opened 2023-11-25
:sparkles: feat: add locale unit test

CoolPlayLin:


__test__/locale.spec.ts
```diff
@@ -0,0 +1,44 @@
...
+import { resolve } from 'node:path'
+import { readdirSync } from 'node:fs'
+import { Language } from '../utils/getLanguage'
+import { includeAllKeys, excludeKeys } from './utils'
+
+const locales = readdirSync(resolve(__dirname, '../locales')).filter((file) => {
+  return file.includes('.json')
+})
+
+describe('should match name regex', () => {
```
cexbrayat, line 11: I'm not sure this test is really useful to be honest. I would remove it.

__test__/utils.ts
```diff
@@ -0,0 +1,48 @@
...
+      return false
+    }
+    if (schema[key] !== null && typeof schema[key] === 'object') {
+      if (!excludeKeys(obj[key], schema[key])) {
+        return false
+      }
+    }
+  }
+  return true
+}
```
cexbrayat, line 48: this file could then be removed

cexbrayat (changes requested):

__test__/locale.spec.ts
```diff
@@ -0,0 +1,28 @@
...
+  }
+  return result;
+}
+
+const locales = readdirSync(resolve(__dirname, '../locales')).filter((file) => {
+  return file.endsWith('.json')
+})
+const defaultKeys = getKeys(en);
+
+describe("should include all keys", () => {
```
cexbrayat, line 22: I would rename the test suite to: `locale files should include all keys`

__test__/locale.spec.ts
```diff
@@ -0,0 +1,28 @@
...
+  locales.forEach((locale) => {
+    it.runIf(!locale.startsWith("en-US"))(`for ${locale}`, () => {
```
cexbrayat, line 24: instead of filtering en-US here, I would remove it earlier and just use `it` here:
```ts
const localesOtherThanEnglish = readdirSync(resolve(__dirname, '../locales')).filter((file) => {
  return file.endsWith('.json') && file !== 'en-US.json'
})
```

cexbrayat (changes requested):
Just a few changes and we're good to merge 👍

CoolPlayLin:
rebase succeeded

cexbrayat (approved):

Issue #393, opened 2023-11-29
The 'build' command in the template/config/typescript/package.json file may contain errors.


mishengqiang:
When I execute the npm command "npm run build -- --mode test," the "@" placeholder only accepts "test," and the "--mode" is missing.
<img width="1092" alt="image" src="https://github.com/vuejs/create-vue/assets/9413032/cf63aed5-ec3c-4ea8-a523-9fe5cc474615">

When I modified the build command to "run-p type-check \"build-only -- {@}\" --" (added -- after build-only), executing the npm command "npm run build -- --mode test" was successful.
<img width="855" alt="image" src="https://github.com/vuejs/create-vue/assets/9413032/1704aefe-80bc-4498-a8c5-eae4d4238583">

mishengqiang:
Sure, thank you for giving me the opportunity to submit a pull request.

9mm:
can somone explain why npm-run-all2 is in package json after a fresh vue-create? what is this even for?

yanhao98:
> can somone explain why npm-run-all2 is in package json after a fresh vue-create? what is this even for?

This npm-run-all package provides 3 CLI commands.

[npm-run-all](https://github.com/bcomnes/npm-run-all2/blob/HEAD/docs/npm-run-all.md)
[run-s](https://github.com/bcomnes/npm-run-all2/blob/HEAD/docs/run-s.md)
[run-p](https://github.com/bcomnes/npm-run-all2/blob/HEAD/docs/run-p.md)


---

then. the lint command uses run-s

```json
{
  "name": "vue-project",
    "lint": "run-s lint:*",
  },
}

```

9mm:
@yanhao98 ahhh interesting, thank you!

One last question... what is @tsconfig/node22 for?

All my old (updated) apps dont have this package, but for last project i decided to do a fresh create-vue. how is this different than previous versions that didnt use it?

IndexXuan:
+1

PR #394, opened 2023-11-30
fix: Correct the 'build' command in the package.json file.

mishengqiang:
The package.json file is located at template/config/typescript/package.json.

```run-p type-check \"build-only {@}\" --```  ->  ```run-p type-check \"build-only -- {@}\" --```

#393

### DIFF
--- a/template/config/typescript/package.json
+++ b/template/config/typescript/package.json
@@ -1,6 +1,6 @@
 {
   "scripts": {
-    "build": "run-p type-check \"build-only {@}\" --",
+    "build": "run-p type-check \"build-only -- {@}\" --",
     "build-only": "vite build",
     "type-check": "vue-tsc --noEmit -p tsconfig.app.json --composite false"
   },


cexbrayat (commented):
It fixes commands like `npm run build -- --sourcemap`, but it no longer works with `pnpm run build --sourcemap` because it adds an extra `--`

@sodatea WDYT?

mishengqiang:
Ah, I didn't consider pnpm and yarn. After testing, yarn performs the best – it passes parameters as expected regardless of whether ```--``` is added or not. npm requires an extra ```--``` to pass parameters as expected. pnpm doesn't need an extra ```--``` to pass parameters as expected.

image

The image below demonstrates when an extra '--' is added.
image

cexbrayat:
Yes, that's why I don't know if we should change the default or not, or maybe there is a way to make it work with all package managers?

sodatea:
Yeah, I think this should be resolved in the upstream `npm-run-all2` package instead.
Before that, maybe we can add a note in the generated README?

ayu-exorcist:
> Yeah, I think this should be resolved in the upstream `npm-run-all2` package instead. Before that, maybe we can add a note in the generated README?

I resolved it in the https://github.com/bcomnes/npm-run-all2/pull/143.

We can run `run-p type-check \"build-only -- {@}\" --` on npm/yarn/pnpm when the PR is merged

ayu-exorcist:
> > Yeah, I think this should be resolved in the upstream `npm-run-all2` package instead. Before that, maybe we can add a note in the generated README?
> 
> I resolved it in the [bcomnes/npm-run-all2#143](https://github.com/bcomnes/npm-run-all2/pull/143).
> 
> We can run `run-p type-check \"build-only -- {@}\" --` on npm/yarn/pnpm when the PR is merged

Already merged in [`npm-run-all2 v6.2.1`](https://github.com/bcomnes/npm-run-all2/releases/tag/v6.2.1)

bcomnes:
Turns out this was a breaking change so I reverted. I need to consider this change a little bit more and at a minimum, put it out in a major release with notes on upgrading.

PR #404, opened 2023-12-11
feat: use type=module in nightwatch projects

cexbrayat:
nightwatch v3.3.4 fixed https://github.com/nightwatchjs/nightwatch/issues/3959

I removed the custom assertions and commands, as they don't work out-of-the-box, they are not used in the generated tests, and they bring little value compared to the official nightwatch documentation


garg3133:
@cexbrayat Just curious what you mean by "they don't work out-of-the-box".

cexbrayat:
The JS version needs to use CJS files whereas the TS version needs to use TS files, and for whatever reason, it failed in some combinations of the generated projects (`typescript-jsx-nightwatch` if I remember correctly). As the custom assertions are not used in the generated tests, I preferred removing them instead of figuring out what we should do to make every combination work. I think people interested in Nightwatch won't have any trouble figuring out how to add custom assertions and commands if they need them.

sodatea (approved):

PR #415, opened 2023-12-28
chore(deps): update script to enable cypress

jasinliu:
Without `--port`, cypress can not visit the default URL. This is confusing.
image

### DIFF
--- a/template/config/cypress/package.json
+++ b/template/config/cypress/package.json
@@ -1,7 +1,7 @@
 {
   "scripts": {
     "test:e2e": "start-server-and-test preview http://localhost:4173 'cypress run --e2e'",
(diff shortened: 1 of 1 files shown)


cexbrayat (commented):
Hi @jasinliu 

Thank you for the pull request.
`--host` is not necessary by default: this is probably because you have a network configuration (security) that prevents cypress from properly working.

I don't think we should include this by default, as it exposes the application on the host.

jasinliu:
> Hi @jasinliu
> 
> Thank you for the pull request. `--host` is not necessary by default: this is probably because you have a network configuration (security) that prevents cypress from properly working.
> 
> I don't think we should include this by default, as it exposes the application on the host.

Thank you very much for your quick reply. I get it.
 
But `--host` did helps for me on Windows11. I'll look into the reasons in my free time.

Issue #246, opened 2023-04-04
Add tailwind as an option


medaminefh:
This is handled here #245

iampapagray:
Any update on this?

medaminefh:
@iampapagray  No no updates here from my side, there is a problem with picking that option (Tailwind) along with Typescript as I mentioned here #245, but it works without Typescript(I've test it before)

iampapagray:
@medaminefh Right. I created a new project with Typescript option. And added Tailwind using the steps here: https://tailwindcss.com/docs/guides/vite#vue . It works fine for now.

PR #416, opened 2023-12-30
feat: add tailwind support

Yordan-Ramchev:
Hi @sodatea @cexbrayat,

I've opened a pull request(re #246) that introduces:
  1. Introduced Tailwind prompt functionality with an untranslated prompt message.
  2. Ensured compatibility for both JavaScript (js) and TypeScript (ts) scenarios.
  3. Modified generateReadme.ts for necessary updates.
  4. Incorporated the Tailwind feature flag into snapshot.mjs.

I would appreciate your time and expertise in reviewing these changes.

Thank you!

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -46,6 +46,7 @@ jobs:
         flag-for-router: ['', '--router']
         flag-for-pinia: ['', '--pinia']
         flag-for-vitest: ['', '--vitest']
+        flag-for-tailwind: ['', '--tailwind']
 
         # It's quite costly to install Cypress & Playwright even with cache.
         # Maybe we can split them into another job so that all the projects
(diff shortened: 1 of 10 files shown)


bmulholland:
One concern is that the approach at https://github.com/vuejs/create-vue-templates is `n!` on the number of options. Probably also the CI matrices? There's already a lot of directories in that repo, and over 100 CI checks... at what point will that approach break down?

I recognize that this is not about your code itself, but about the practicalities of maintaining all combinations of options.

At a minimum, you probably need to add the tailwind option to https://github.com/vuejs/create-vue/blob/main/.github/workflows/ci.yml

Yordan-Ramchev:
@bmulholland thank you for the feedback. I have added tailwind flag in CI and that's a reasonable concern.

cexbrayat:
Note: this PR is a duplicate of an oldest PR #245 If one of the two is merged, the other can be closed.

I personally don't think it's worth the added complexity to add tailwind support, but I'll let @sodatea decide.

Issue #422, opened 2024-01-09
ESlint plugin playwright


Yordan-Ramchev:
**Is your feature request related to a problem? Please describe.**

> In the long run we need to add https://github.com/playwright-community/eslint-plugin-playwright/ to the playwright setup too.

See more: https://github.com/vuejs/create-vue/commit/f280b26fe938080967c332d88fa6eaba70dab870

**Describe the solution you'd like**
Incorporate [eslint-plugin-playwright](https://github.com/playwright-community/eslint-plugin-playwright) as a development dependency. When opting for the End-to-End (E2E) Playwright solution, ensure extension of the ESLint configuration.

**Additional context**
1. Add [eslint-plugin-playwright](https://github.com/playwright-community/eslint-plugin-playwright) in /template/eslint/package.json
2. Update index.ts > pass playwright option to renderEslint function
3. Add logic for extending eslint with eslint-plugin-playwright plugin 
    `"extends": ["plugin:playwright/recommended"]`

PR #433, opened 2024-01-26
feat: eslint plugin playwright

georgi-mateev:
Re #422 
- [x] Add [eslint-plugin-playwright](https://github.com/playwright-community/eslint-plugin-playwright) in /template/eslint/package.json
- [x] Update index.ts > pass playwright option to renderEslint function
- [x] Add logic for extending eslint with eslint-plugin-playwright plugin
"extends": ["plugin:playwright/recommended"]

### DIFF
--- a/utils/renderEslint.ts
+++ b/utils/renderEslint.ts
@@ -13,7 +13,7 @@ const eslintDeps = eslintTemplatePackage.devDependencies
 
 export default function renderEslint(
   rootDir,
-  { needsTypeScript, needsCypress, needsCypressCT, needsPrettier }
+  { needsTypeScript, needsCypress, needsCypressCT, needsPrettier, needsPlaywright }
 ) {
   const additionalConfig: Linter.Config = {}
   const additionalDependencies = {}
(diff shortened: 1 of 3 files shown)


Yordan-Ramchev:
Nice one @georgi-mateev 
Could you guys provide us with feedback @sodatea @cexbrayat

utils/renderEslint.ts
```diff
@@ -35,6 +35,11 @@ export default function renderEslint(
     additionalDependencies['eslint-plugin-cypress'] = eslintDeps['eslint-plugin-cypress']
   }
 
+  if (needsPlaywright) {
+    additionalConfig.extends = ['plugin:playwright/recommended']
```
cexbrayat, line 39: shouldn't the plugin apply only the Playwright files (as done for Cypress above)?

cexbrayat (changes requested):

utils/renderEslint.ts
Yordan-Ramchev: Yes, ESLint will apply the rules defined in the root of your configuration to all files in your project (which is not ideal). 
@georgi-mateev Can you target **e2e** folder - something like `files: ['e2e/**/*.{js,ts,jsx,tsx}']`?

georgi-mateev:
@cexbrayat,  @Yordan-Ramchev  Thanks for the feedback and the help, appreciate it!

### Changes
Plugin now only applies to specified files - 
eslint-plugin

Regarding `files` we think its versatile enough to target files having either `.test.` or `.spec.` signature both in the the main folder and its subfolders, if any.

Yordan-Ramchev (approved):
Looking good

cexbrayat (approved):
LGTM thanks @georgi-mateev 👍

PR #436, opened 2024-01-30
fix: git sees .vscode folder even if in .gitignore

### LINKED ISSUE #435: git considers .vscode/ folder after new projects even if in the .girignore file
# Description
After I created a new project and initialized a repository when running `git status` git shows you the `.vscode` folder, even though it is in the .gitignore file.

# What I did

I ran `npm create vue@latest` in a new terminal, entered a name and pressed Enter many times to skip everything.
Next:
```bash
cd test
npm install
git init
git status
```
and the terminal output was:

```bash
On branch master

No commits yet

Untracked files:
  (use "git add <file>..." to include in what will be committed)
	.gitignore
	.vscode/
	README.md
	index.html
	jsconfig.json
	package-lock.json
	package.json
	public/
	src/
	vite.config.js

nothing added to commit but untracked files present (use "git add" to track)
```

You can see `.vscode/` is there.

# How to fix
Open the .gitignore file with your preferred editor and change `.vscode/*` to `.vscode/`.

kada49:
Fixes #435 (Hopefully the right way (: )

### DIFF
--- a/template/base/_gitignore
+++ b/template/base/_gitignore
@@ -18,7 +18,7 @@ coverage
 /cypress/screenshots/
 
 # Editor directories and files
-.vscode/*
+.vscode/
 !.vscode/extensions.json
 .idea
 *.suo


cexbrayat (commented):
This is because we want to version some files inside the .vscode directory, like `extensions.json`, so we need to keep the asterisk

Issue #168, opened 2022-09-27
Replace `minimist` with `@pkgjs/parseargs`


sodatea:
https://github.com/substack/minimist now returns 404
https://twitter.com/dan_abramov/status/1573715996573155336
Likely because @substack didn't wanna use 2FA.

Anyway, Node.js provides a built-in `util.parseArgs` helper since v18.3.0, which uses https://github.com/pkgjs/parseargs underlylingly. The package supports Node.js >= 14. We should use it instead.

broql:
Is this issue still valid? Can I take it?

sodatea:
Yeah, go ahead please.

PR #443, opened 2024-02-05
refactor: minimist to parseargs

Yordan-Ramchev:
# Pull Request

## Description
1. Remove minimist package from package.json
2. index.ts > Migrate minimist related code to parseArgs
3. Delete minimist from LICENSE.

## Related Issue
re #168 

## Checklist

- [x] All existing tests pass
- [x] Code comments added/updated

## Notes for Reviewers
1. **alias** is not supported by **parseArgs**. It has `short {string}` A single character alias for the option, but it's not matching our case.
2. **strict** is set to **false** because it will trigger a console error when an argument is passed that is not defined in the options.


cexbrayat (approved):
LGTM, but I'll let @sodatea decides if he wants this or not.

Note: `parseArgs` was added in v16.17 and v18.3, but is experimental until v20 (https://nodejs.org/api/util.html#utilparseargsconfig)

sodatea (approved):
Thanks for the PR! It's really helpful.

Considering the `parseArgs` API is already stable in Node.js v20, I think it's safe to assume no breaking changes would ship to v18.x, so let's use it.

PR #448, opened 2024-02-21
chore: remove unused parameter

btea:


cexbrayat (commented):
Good catch, but maybe the proper fix is to use the parameter to generate the readme?

btea:
Makes sense, but I'm not sure what the generated content based on the `needsNightwatchCT` parameter should be?

cexbrayat:
I think `needsNightwatch` is used twice, instead I think it should be `needsNightwatchCT` here https://github.com/vuejs/create-vue/pull/448/files#diff-a4daf76507373a095f626fff68f5cecaa33d5d10a4dbca6aaad95ef30a80bd30R125

btea:
Okay, I have made the modifications.

cexbrayat (approved):

PR #450, opened 2024-02-21
feat: use `Intl.getCanonicalLocales` check locale

btea:
When users configure LANG related information by themselves, irregular spelling or capitalization problems may occur. However, I am not sure whether to throw an error directly to terminate the process at this time or just output the error message and continue to perform the following steps. (The current action I take is to print the error message and continue.)

Related commit https://github.com/vuejs/create-vue/commit/bb11f04a5319cecde4bd6e87c3bbb1612d7a88ae

### DIFF
--- a/index.ts
+++ b/index.ts
@@ -61,6 +61,9 @@ function emptyDir(dir) {
 }
 
 async function init() {
+  // The language information obtained may be incorrect. In this case, an error should be thrown and the banner information should not be printed.
+  const language = getLanguage()
+
   console.log()
   console.log(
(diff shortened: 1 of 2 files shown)


sodatea (approved):
Thanks!

Note to myself: the `@ts-ignore` comment can be removed after upgrading to TypeScript 5.4

https://github.com/microsoft/TypeScript/pull/56079

bmulholland:
In future, ts-expect-error is a good alternative to ts-ignore; it'll raise an error when it's no longer needed, so you don't have to remember :)

PR #471, opened 2024-03-18
fix: Command description for Bun

ccjr1120:
I saw the command to create Vue using bun on Vuejs.org, but the description after creation is still stuck at npm🙏


cexbrayat (approved):
Thanks @ccjr1120

PR #479, opened 2024-03-27
i18n: Update zh-Hant.json

toto6038:
Refine translation for Traditional Chinese by using glossaries used in Taiwan.


sodatea (approved):
Thanks!

Issue #486, opened 2024-04-05
The e2e tests in the subfolders under e2e/ are being executed by vitest


messenjer:
The end-to-end tests located in the subfolders under e2e are being executed by vitest because they are not properly excluded.

```
e2e/test.spec.ts
e2e/frontend/login.spec.ts. <-- this playwright test is executed by vitest and failed
```

The wildcard for exclude in the vitest config should be `**` instead of `*` and a related issue here : https://github.com/vitest-dev/vitest/issues/3102 

```diff
-      exclude: [...configDefaults.exclude, 'e2e/*'],
+      exclude: [...configDefaults.exclude, 'e2e/**'],
```

Here is the PR : https://github.com/vuejs/create-vue/pull/487

PR #487, opened 2024-04-05
fix: exclude from vitest the playwright e2e/** subfolders

messenjer:
Fix : https://github.com/vuejs/create-vue/issues/486


cexbrayat (approved):
Good catch, thanks @messenjer 👍

PR #491, opened 2024-04-10
fix: typo

gyhyfj:


sodatea:
Not a typo.
This is a [possessive apostrophe](https://en.wikipedia.org/wiki/Apostrophe#Typographic_form).

According to Unicode, [U+2019 is the preferred character to use for apostrophe](https://arc.net/l/quote/ozqtpemn).

PR #514, opened 2024-05-08
fix: write merged setting config

Muluk-m:
### Description

  Write merged setting config when render template


btea (approved):

cexbrayat (approved):
Good catch, thanks

PR #528, opened 2024-05-29
fix: use en quote replace cn quote

shiux-lab:


sodatea:
https://github.com/vuejs/create-vue/pull/491#issuecomment-2046893479

PR #540, opened 2024-07-04
fix(playwright): use `npm run` to start the server on default port

zetaraku:
### Description

When running Playwright end-to-end testing from [Playwright Test for VSCode](https://marketplace.visualstudio.com/items?itemName=ms-playwright.playwright), an error is thrown:

```
Error: Process from config.webServer was not able to start. Exit code: 1
```

This is probably because the extension runs without adding `node_modules/.bin` to `PATH` like `npm run` does.

### What this PR does

This PR uses `npm run` to start the server and adjusts the `port` according to the default port of `vite dev` or `vite preview`.


cexbrayat (approved):
LGTM, thanks for the PR!

Issue #545, opened 2024-07-23
i18n vue plugin option ?


ArcadeCode:
### Please describe

I use [vue I18n](https://vue-i18n.intlify.dev/) who is really helpful for Internationalization but he need some configuration and I will really appreciate if I can have an option in create-vue for this, but I checked the code and I'm not qualified to understand it.
### Additional context

To make a fast configuration the vue-i18n doc propose to config all in the main.ts but it's more clear to make a local(s) folder with all traductions and import it in the main.ts or use a i18n folder: 
![image](https://github.com/user-attachments/assets/4f7ab16b-2ad3-48bd-bb3a-eddd804a7bf7)

Anyone think it will be good idea ?

Procrustes5:
I am developing in Japanese and Korean, totally agree with you.

When the project grows to a certain extent, I have often thought that it would be nice if the i18n was set up at the stage of making the project.

I think it's also good for the configuration of creating folders on i18n.

Procrustes5:
https://github.com/vuejs/create-vue/pull/548
Made a pull request based on your issue. If you have any opinions, please give it to me.

messenjer:
In general, I put my i18n configuration into a separate folder: `plugins/i18n.ts`

I keep the translation files in the locales folder: `src/locales/en.json`.

This separation makes it easy to specify the folder containing the translations when using various translation tools:

* The VSCode extension: https://github.com/lokalise/i18n-ally

```
{
    "i18n-ally.localesPaths": [
        "src/locales"
    ]
}
```

* CI/CD : https://lokalise.com/product/apps/code-repositories/github 
* The Vite plugin for vue-i18n: https://github.com/intlify/bundle-tools/blob/main/packages/unplugin-vue-i18n/README.md 

```
export default defineConfig({
  plugins: [
    VueI18nPlugin({
      include: [path.resolve(__dirname, './src/locales/**')],
    }),
  ],
})
```

I find this approach to be more modular.

ArcadeCode:
Yes it will be more modular in this way, but in this idea why don't moving router.ts in the plugins folder ?

messenjer:
In the previous vue installer `vue-cli-i18n` https://github.com/intlify/vue-cli-plugin-i18n 

The file was located here : 

`src/i18n.(js|ts)`

And the default localeDir was `src/locales`

it should be another solution.

ArcadeCode:
Sounds good to me if we take the example from the i18n vue-cli plugin.

[...]

PR #548, opened 2024-08-02
feat: create with vue-i18n

Procrustes5:
### Description

issue : https://github.com/vuejs/create-vue/issues/545

Added vue-i18n to the option. I made it with the directory structure suggested in the above issue.


.idea/vcs.xml
```diff
@@ -0,0 +1,7 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<project version="4">
+  <component name="VcsDirectoryMappings">
+    <mapping directory="" vcs="Git" />
+    <mapping directory="$PROJECT_DIR$/playground" vcs="Git" />
+  </component>
+</project>
```
cexbrayat, line 7: can you remove all the `.idea` files that you added to this commit?

template/config/vue-i18n/src/i18n/index.js
```diff
@@ -0,0 +1,15 @@
...
+
+const messages = {
+  en: {
+    // English Transtlations
+  }
+}
+
+export default createI18n({
+  legacy: false,
+  globalInjection: true,
```
cexbrayat, line 11: I thnink this is the default value, so the line can be removed

template/config/vue-i18n/src/i18n/index.js
```diff
@@ -0,0 +1,15 @@
...
+  locale: 'en',
+  fallbackLocale: 'en',
+  messages
+})
```
cexbrayat, line 15: Also, when used in TS, there are additional informations that should be added to have a proper typechecking:
https://vue-i18n.intlify.dev/guide/advanced/typescript.html

cexbrayat (commented):
This is a good start.

There should be at least one snapshot test to check that everything works.
Ideally, the should introduced a modified component template to use a translation that could be checked in a test.

Procrustes5:
@cexbrayat 
Thank you for review!

> Ideally, the should introduced a modified component template to use a translation that could be checked in a test.

I've got it !
Is it okay to do minimal component testing first, and then plan to test all possible cases later? Of course, I want to keep responding continuously.

Procrustes5:
@cexbrayat 
https://github.com/vuejs/create-vue-templates/pull/4
I wrote snapshot test case. I'd appreciate it if you could check this for me.

cexbrayat:
@Procrustes5 Sorry, I should have explained: to add a snapshot test, you don't have to open a PR on create-vue-templates, you have to update the `snaphot.mjs` script in this PR. `create-vue-templates` is then automatically updated when we release a new version of create-vue.

Procrustes5:
@cexbrayat 
Thank you!
I've changed snapshot. plz check this!

scripts/snapshot.mjs
```diff
@@ -15,7 +15,8 @@ const featureFlags = [
   'vitest',
   'cypress',
   'playwright',
-  'nightwatch'
+  'nightwatch',
+  'vue-i18n'
```
cexbrayat, line 19: I think it should be removed from here because it's going to generate too many snapshots

template/tsconfig/vue-i18n/i18n.d.ts
```diff
@@ -0,0 +1,11 @@
...
+ * you need to import the some interfaces
+ */
+import { DefineLocaleMessage } from 'vue-i18n'
+
+declare module 'vue-i18n' {
+  // define the locale messages schema
+  export interface DefineLocaleMessage {
+    hello: string
+  }
+}
```
cexbrayat, line 11: I don't think this file should be necessary? With TypeScript, you need to do something like:

```ts
import { createI18n } from 'vue-i18n'
import enUS from './locales/en-US.json'

// Type-define 'en-US' as the master schema for the resource
type MessageSchema = typeof enUS

const i18n = createI18n<[MessageSchema], 'en-US'>({
  locale: 'en-US',
  messages: {
    'en-US': enUS
  }
})
```

template/tsconfig/vue-i18n/tsconfig.i18n.json
```diff
@@ -0,0 +1,9 @@
+{
+  "extends": "../base/tsconfig.app.json",
+  "compilerOptions": {
+    "composite": true,
+    "tsBuildInfoFile": "./node_modules/.tmp/tsconfig.vue-i18n.tsbuildinfo",
+    "types": ["i18n"]
+  },
+  "include": ["src/template/config/vue-i18n/**/*", "src/template/config/vue-i18n/**/*.json"]
+}
```
cexbrayat, line 9: And i think this file can be removed as well?

Procrustes5:
@cexbrayat 
Thank you for review!
You mean `template/tsconfig/vue-i18n` is unnecessary, is it correct?
I have changed, please check it one more time!

cexbrayat:
@Procrustes5 
As I was pointing out, when using `vue-i18n`, the generated `src/i18n/index.ts` file should look like this I think:
```ts
import { createI18n } from 'vue-i18n'
import enUS from './locales/en-US.json'

type MessageSchema = typeof enUS

const i18n = createI18n<[MessageSchema], 'en-US'>({
  locale: 'en-US',
  messages: {
    'en-US': enUS
  }
})
```

Also, as a second step, the generated HelloWorld component could show the usage of `t()` (for example https://github.com/vuejs/create-vue/blob/main/template/code/default/src/components/HelloWorld.vue could have `You’ve successfully created a project with` translated). And the generated unit test could check this to ensure the addition of `vue-i18n` works well.

cexbrayat:
Oh and I forgot something: the Pr should also update `ci.yml` to actually test the `vue-i18n` flag.
We don't want to test with all combinations, but maybe you could do the same that was done for the `devtools` flag.

Procrustes5:
@cexbrayat 
Thank you so much for the easy explanation. I've modified it, so please check it out!

template/code/typescript-i18n/src/App.vue
```diff
@@ -0,0 +1,47 @@
...
+    margin: 0 2rem 0 0;
+  }
+
+  header .wrapper {
+    display: flex;
+    place-items: flex-start;
+    flex-wrap: wrap;
+  }
+}
+</style>
```
cexbrayat, line 47: Is this file necessary? If it doesn't change from the `typescript-default` versions, then I think it should not be duplicated if possible (same for the other components).

template/code/typescript-i18n/src/components/HelloWorld.vue
```diff
@@ -0,0 +1,45 @@
...
+}>()
+
+const { t } = useI18n()
+</script>
+
+<template>
+  <div class="greetings">
+    <h1 class="green">{{ msg }}</h1>
+    <h3>
+      {{ t('greetings.message') }}
```
cexbrayat, line 15: 👍

cexbrayat (commented):
Nice work. I'm just wondering if the components which are the same in the `typescript-default` template could not be removed to avoid duplication. Other than that, looks good to me.

template/code/typescript-i18n/src/App.vue
Procrustes5: @Procrustes5 
Oh, I'd not understood this architecture yet. Thank you
modified now!

Procrustes5:
@cexbrayat 
It has been left here for a long time in a state where I can't proceed with anything. Is there anything I can do?

Issue #437, opened 2024-01-31
type-check reports errors twice


unshame:
Creating a project with typescript and vitest

```
Vue.js - The Progressive JavaScript Framework

√ Add TypeScript? ... Yes
√ Add Vitest for Unit Testing? ... Yes
```

making a mistake somewhere and then running `type-check` script reports the errors twice:

![image](https://github.com/vuejs/create-vue/assets/18334400/4ad0efc9-763d-4b4e-bd55-d57ac401388e)

This is due to the fact that "src" is included in both `tsconfig.app.json` and `tsconfig.vitest.json`. This was introduced with https://github.com/vuejs/create-vue/pull/274

Not sure how to fix this. It's nice to have a separate tsconfig for vitest, but not sure it's worth it if it double reports the errors because of it.

cexbrayat:
Hi @unshame , thanks for the report.

I noticed it, but did not really mind personally.

@sodatea do you have an idea on how we could do better for this?

sjihdazii:
> Creating a project with typescript and vitest
> 
> ```
> Vue.js - The Progressive JavaScript Framework
> 
> √ Add TypeScript? ... Yes
> √ Add Vitest for Unit Testing? ... Yes
> ```
> 
> making a mistake somewhere and then running `type-check` script reports the errors twice:
> 
[...]

PR #549, opened 2024-08-04
Feat project references

unshame:
### Description

Fixes duplicate typechecking and error reporting when running type-check with vitest or cypress component tests enabled.

Adds usage of project references to vitest and cypress component test tsconfigs. Adds declaration emit to the tsconfig.app.json. Only includes test files in unit testing tsconfigs.

Fixes https://github.com/vuejs/create-vue/issues/437

#### E2E tests

I didn't touch tsconfigs for e2e tests. They most likely can be setup the same way as unit tests, I don't see why they need to be in a separate folder. There are more changes that need to be made to achieve this though, that should be done in a separate PR. 

The nightwatch config is the weirdest case - there is just a copy of tsconfig.app.json in it that overwrites the original one. I tried to set it up to work the same way as cypress-ct does, but looks like the vue plugin doesn't have any types, so it was giving me errors. I deleted that instead of copying the new content of tsconfig.app.json into it.


template/tsconfig/vitest/tsconfig.vitest.json
```diff
@@ -1,11 +1,18 @@
 {
   "extends": "./tsconfig.app.json",
```
unshame, line 2: I kept this extending the app config, but maybe it's better to make it extend @vue/tsconfig/tsconfig.dom.json directly? That's how I've set it up in my project, makes it easier to manage.

template/tsconfig/cypress-ct/tsconfig.cypress-ct.json
```diff
@@ -1,15 +1,17 @@
...
-    "src/**/*",
-    "src/**/*.vue",
+    "src/**/__tests__/*",
     "cypress/support/component.*",
     "cypress/support/commands.ts"
   ],
   "exclude": [],
   "compilerOptions": {
     "composite": true,
+    "outDir": "./node_modules/.tmp/cypress-ct",
```
unshame, line 11: It's not strictly necessary to have the declarations output for anything but the tsconfig.app.json right now, but I think it makes sense to make them all work the same way.

cexbrayat:
Let's hear from @sodatea what he thinks about this setup

template/tsconfig/vitest/tsconfig.vitest.json
codethief: Yeah, I'd argue it's better to keep production and test configs separate (and rather have them extend a common base config), so that people won't e.g. add types to `tsconfig.app.json` that don't belong in test files.

template/tsconfig/cypress-ct/tsconfig.cypress-ct.json
codethief: > It's not strictly necessary to have the declarations output for anything but the tsconfig.app.json right now

Note that this will change in TypeScript 5.6, see https://github.com/microsoft/TypeScript/issues/32651#issuecomment-2260378908

template/tsconfig/vitest/tsconfig.vitest.json
```diff
@@ -1,11 +1,18 @@
...
+  "include": [
+    "src/**/__tests__/*"
+  ],
   "exclude": [],
   "compilerOptions": {
     "composite": true,
+    "outDir": "./node_modules/.tmp/vitest",
     "tsBuildInfoFile": "./node_modules/.tmp/tsconfig.vitest.tsbuildinfo",
```
codethief, line 10: tsBuildInfoFile can be removed since now that we set outDir, its default value will be <outDir>/<config name>.tsbuildInfo, compare https://www.typescriptlang.org/tsconfig/#tsBuildInfoFile

codethief:
Thanks so much for picking this up, @unshame ! :)

unshame:
Interesting. 

I do see the error in vscode:
image

Webstorm on the other hand has no issues with it and it's using the same version of vue language server.
image
image

I also don't have issues with vut-tsc, it reports the errors correctly and doesn't infer the type of the component as any:
image

So looks like a vscode extension bug?

codethief:
@unshame Interesting. My coworker has been reporting issues at least in IntelliJ, though. Would you mind re-posting your comment in https://github.com/vuejs/language-tools/issues/3526 together with the exact versions of vue-tsc, VSCode extension, WebStorm, WebStorm Vue extension etc.?

codethief:
@messenjer vue-tsc is still at 2.0.29, see https://www.npmjs.com/package/vue-tsc . I don't know which Volar version vue-tsc 2.0.29 uses but it was released ~a month ago. (There are no release notes on https://github.com/vuejs/language-tools yet.)

Meanwhile, @vuejs/language-tools master [updated to Volar 2.4.0 only 2 days ago](https://github.com/vuejs/language-tools/commit/d24c984d13889518f0404a8f673ab6c0c2ca0e70). So we'll have to wait for a new vue-tsc version that uses Volar 2.4.0.

codethief:
Since my last comment a lot has happened: [One ticket](https://github.com/vuejs/language-tools/issues/4711) has been closed, [another (new) ticket](https://github.com/vuejs/language-tools/issues/4750) is still open. Personally, I decided the remaining issue is no longer a blocker, since it only affects IDE support in minor ways and type checking is not affected. So in our project I went ahead and implemented project references throughout and until now I haven't received any major complaints from my coworkers.

For create-vue the considerations might be different, though, I don't know.

PR #554, opened 2024-08-08
fix: use child_process module to create project

btea:
### Description

The following error will occur on Windows platform when executing `pnpm snapshot`.

QQ_1723075297659

### DIFF
--- a/scripts/snapshot.mjs
+++ b/scripts/snapshot.mjs
@@ -1,4 +1,5 @@
 #!/usr/bin/env zx
+import { execSync } from 'node:child_process'
 import fs from 'node:fs'
 import 'zx/globals'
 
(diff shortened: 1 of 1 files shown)


cexbrayat (commented):
I'm not sure we should change this. `zx` is explicit about having to use bash even on Windows (see https://google.github.io/zx/setup#bash).

btea:
Yeah, I noticed this too. However, the default terminal launched by vscode is not bash, and it needs to be switched manually, which seems a bit troublesome. Or, is there any other way to use base to run files by default? 🤔

cexbrayat:
@btea I don't know how you can switch on VS Code, but I guess there is a way to define your default terminal. Let's close then

btea:
I found that we use `npm-run-all2`, but I think we can directly replace it with `pnpm run build && pnpm run snapshot`, maybe we can remove this dependency. What do you think? @cexbrayat 
https://github.com/vuejs/create-vue/blob/f283ab4fdc0cbef48e8ecde77d52f2398540e08d/package.json#L23

cexbrayat:
@btea Sure, why not

PR #557, opened 2024-08-14
ci: add eslint check on some jobs

cexbrayat:
### Description

This adds the --eslint flag on some jobs and checks there are no warnings

PR #558, opened 2024-08-14
test: add unit test for eslint configs

cexbrayat:
### Description

The commit adds a unit test for the eslint configurations and properly formats our tests.

On top of #557


cexbrayat:
merged as part of #559

PR #559, opened 2024-08-14
feat: add @vitest/eslint-plugin when using vitest 

cexbrayat:
### Description

The plugin is now officillay recommended, so let's add it to the eslint config when selecting both eslint and vitest.

On top of #558


btea (approved):
LGTM.
It is better for you to merge these three PRs in order.

PR #575, opened 2024-10-09
test: re-add eslint unit tests

cexbrayat:
### Description

Also fixes tiny discrepancies:
- all files are now declared with single quotes
- dot notation is preferred to access configs when possible


haoqunjiang (approved):
Thanks! Esp. the single quote issue… I totally overlooked that one.

PR #582, opened 2024-10-15
chore: delete kolorist as the project now uses kleur/colors

Yordan-Ramchev:
### Description

Remove `kolorist` from package.json as the project now uses `kleur/colors`.


cexbrayat (approved):
👍

PR #594, opened 2024-10-30
chore: mark template files as CC0

btea:
### Description
Excerpt from https://github.com/vitejs/vite/pull/18366


cexbrayat (approved):
LGTM but I'll let @haoqunjiang review this

haoqunjiang (approved):
Based on my research, I think [MIT-0](https://opensource.org/license/MIT-0) or [0BSD](https://opensource.org/license/0bsd) might be more desirable for users, as the concerns raised by [Free Software Foundation](https://www.gnu.org/licenses/license-list.html#CC0).
But there isn't an imminent downside of CC0 right now; no users have ever complained about it. So, let's keep it in sync with vite until further notice.

Thanks for the PR!

PR #630, opened 2024-12-05
[feat] Added support for Bun

mchao123:
Added support for Bun, added runtime selectors.

This is a tentative change that may require a review of the code.

### DIFF
--- a/index.ts
+++ b/index.ts
@@ -63,6 +63,22 @@ function emptyDir(dir) {
   )
 }
 
+// get now user runtime
+function detectRuntime() {
+  // @ts-ignore
+  if (typeof Bun !== 'undefined') {
+    return 'bun'
+  }
+  if (typeof process !== 'undefined' && process.versions && process.versions.node) {
+    return 'node'
+  }
+  // @ts-ignore
+  if (typeof Deno !== 'undefined') {
+    return 'deno'
+  }
+  return 'unknown'
+}
+
 const helpMessage = `\
 Usage: create-vue [FEATURE_FLAGS...] [OPTIONS...] [DIRECTORY]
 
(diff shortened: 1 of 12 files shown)


cexbrayat:
@mchao123 Thanks for the PR

@haoqunjiang Is this something we want to support as an option and yet another feature to maintain? Or could we just add a note in the README instead (as, IIUC, this just requires users to use a specific tsconfig file?).

haoqunjiang:
> @haoqunjiang Is this something we want to support as an option and yet another feature to maintain? Or could we just add a note in the README instead (as, IIUC, this just requires users to use a specific tsconfig file?).

I'm open to add bun support considering it's low-effort. But I don't think we need yet another option to show to every user. User-agent detection should suffice.

mchao123:
> > @haoqunjiang Is this something we want to support as an option and yet another feature to maintain? Or could we just add a note in the README instead (as, IIUC, this just requires users to use a specific tsconfig file?).
> 
> I'm open to add bun support considering it's low-effort. But I don't think we need yet another option to show to every user. User-agent detection should suffice.

Some people prefer to use yarn or pnpm to manage their project creation.

haoqunjiang:
> Some people prefer to use yarn or pnpm to manage their project creation.

But I don’t think that’s a common enough use case for us to support.

mchao123:
It's been a long time has there been an update on this?I always need to do some extra work after creating a project in order to adapt to bun, it would be great if this was supported at creation time, thank you guys.

mchao123:
> @mchao123 I think the user-agent detection proposition from @haoqunjiang is a good one: not too much maintenance work, and not _another_ prompt for users. Feel free to ping us when you'll have reworked the PR 👍


Ok, I'll close this PR and finish this modification based on the latest branch again.

mchao123:
I've made the change to show the runtime checkbox only when the user carries the parameter "--runtime".Maintaining a proprietary version of a package only requires creating a new directory in the others-runtime folder.

mchao123:
> 我认为 user-agent detection 的提议很好：没有太多的维护工作，也没有_给用户另一个_提示。当您重新编写 PR 👍 时，请随时 ping 我们

Help reviewing the code, thanks! @cexbrayat

PR #635, opened 2024-12-12
fix: remove `composite: true` from tsconfigs

haoqunjiang:
### Description

We are only using `references` in a solution-style tsconfig. According to discussions at https://github.com/microsoft/TypeScript/issues/60465, such usage doesn't require `composite: true` to be set in sub-configs.

Removing this field loosens the constraints on these configs that all checked files to be explicitly listed in `files` or `includes`.

After this change, type errors in source code would only be reported twice if they're also imported by unit test specs, in contrast to always be reported twice prior to the change.
I know this is not ideal yet, but it's still an improvement, and might help catch some edge cases such as https://github.com/vuejs/create-vue/issues/437#issuecomment-2468910155

In the long run, we should still keep an eye on https://github.com/vuejs/create-vue/pull/549 (pending https://github.com/vuejs/language-tools/issues/4750). Cross-referencing might be a more intuitive configuration, and should be the desirable configuration if we opt into Vitest Browser Mode.

### DIFF
--- a/template/tsconfig/base/tsconfig.app.json
+++ b/template/tsconfig/base/tsconfig.app.json
@@ -3,7 +3,6 @@
   "include": ["env.d.ts", "src/**/*", "src/**/*.vue"],
   "exclude": ["src/**/__tests__/*"],
   "compilerOptions": {
-    "composite": true,
     "tsBuildInfoFile": "./node_modules/.tmp/tsconfig.app.tsbuildinfo",
 
     "paths": {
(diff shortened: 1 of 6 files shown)


cexbrayat (approved):

charnog:
@haoqunjiang to keep you informed about the implications of this PR.

[There was a "Heisenbug" issue](https://github.com/vuejs/language-tools/issues/1232) about `Default export of the module has or is using private name 'Props'` TypeScript error ([demo repo](https://github.com/charnog/vue-default-export)). It was resolved by removing `composite: true`. More info [here](https://github.com/vuejs/language-tools/issues/1232#issuecomment-2601302894).

Issue #2, opened 2021-08-27
Feedback (Aug 27)


yyx990803:
## File Structure

- [x] Rename `vite-env.d.ts` to just `env.d.ts` since it's no longer vite-specific
- [x] Remove `jsconfig.json` and `env.d.ts` for non-TS projects (no longer necessary for latest Volar)

## Welcome Screen

- [x] Community should mention the @vuejs twitter account
- [x] The welcome screen doesn't need the explicit scheme switch - it should just follow system preference.
- [x] It feels a bit messy to have all these components/composables for the welcome screen in a fresh project. We should move them all under a `src/welcome` directory so the user can easily delete them. `src/App.vue` should just import and render a `src/welcome/Welcome.vue` component.
- [x] Make the wording the for the Documentation item more succint: "Vue’s official documentation provides you with all the information you need to get started."

## Publishing

- [x] Move all deps to `devDependencies` +  bundle with `esbuild` before publishing to reduce network requests

PR #636, opened 2024-12-12
feat: use a `--bare` flag to generate a template without too much boilerplate

### LINKED ISSUE #112: Add option for bare install
The Vue CLI provided an option -b to exclude the scaffolded components. It would be nice if there was a way to do this with create-vue either through a flag or as one of the options.

### LINKED ISSUE #186: Question: How to start with empty template?
I want the same thing, but minus all of the starter components and css.

### LINKED ISSUE #300: Created project is way too complex, not an actual scaffold
How is this officially "The recommended way to start a Vite-powered Vue project"?! It does create such a project, but before I can _start_ coding my own actual project, I have to delete a ton of garbage: Components I don't need, CSS I don't need, logos and icons I don't need...

In its current state, this project creates a good example for beginners that shows off a few of Vue's features and gives links for further reading. But it massively fails at just providing an empty template from which to **start one's own project.** I was expecting a scaffold, which means the **minimal setup** to get it to run and say something like "Hello World", nothing more.

Like @ZackPlauche said it in #186: "I want the same thing, but minus all of the starter components and css." I opened a new issue because
1. over the past 8 months that issue didn't get any attention (except, notably, five thumbs-up reactions of people thinking the same) and
2. he titled his issue a "question", whereas I see mine more as a bug report, in the sense of "this software does not do what it is supposed to do, please fix".

haoqunjiang:
Closes #112
Closes #186
Closes #300
Closes #637

~~The minimal template is a cut-down version of the default template;
therefore, the `--minimal` flag is mutually exclusive with all other flags.~~

I chose not to include it in the prompts because it's a very specific
use case, and it's not a good idea to clutter the CLI with too many
options that are not commonly used.

~~I didn't design it as an add-on because the logic would be too complex,
and the use case would be even more niche.~~

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -51,7 +51,7 @@ jobs:
           verification-script:
             - pnpm --filter "\!*typescript*" build
             - pnpm --filter "*typescript*" build
-            - pnpm --filter "*vitest*" test:unit
+            - pnpm --filter "*vitest*" --filter "\!*bare*" test:unit
             - pnpm --filter "*eslint*" lint --no-fix --max-warnings=0
             - pnpm --filter "*prettier*" format --write --check
             # FIXME: it's failing now
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -49,13 +49,13 @@ jobs:
           node-version: [18, 20, 22]
           os: [ubuntu-latest, windows-latest, macos-latest]
           verification-script:
-            - pnpm --filter "\!*typescript*" build
-            - pnpm --filter "*typescript*" build
-            - pnpm --filter "*vitest*" --filter "\!*bare*" test:unit
-            - pnpm --filter "*eslint*" lint --no-fix --max-warnings=0
-            - pnpm --filter "*prettier*" format --write --check
+            - pnpm --filter '!*typescript*' build
+            - pnpm --filter '*typescript*' build
+            - pnpm --filter '*vitest*' --filter '!*bare*' test:unit
+            - pnpm --filter '*eslint*' lint --no-fix --max-warnings=0
+            - pnpm --filter '*prettier*' format --write --check
             # FIXME: it's failing now
-            # - pnpm --filter "*with-tests*" test:unit
+            # - pnpm --filter '*with-tests*' test:unit
       runs-on: ${{ matrix.os }}
       continue-on-error: ${{ matrix.os == 'windows-latest' }}
       env:
@@ -163,11 +163,11 @@ jobs:
 
       - name: Run build script
         working-directory: ./playground
-        run: pnpm --filter "*${{ matrix.e2e-framework }}*" build
+        run: pnpm --filter '*${{ matrix.e2e-framework }}*' build
 
       - name: Run e2e test script
         working-directory: ./playground
-        run: pnpm --filter "*${{ matrix.e2e-framework }}*" --workspace-concurrency 1 test:e2e
+        run: pnpm --filter '*${{ matrix.e2e-framework }}*' --workspace-concurrency 1 test:e2e
 
       - name: Cypress component testing for projects without Vitest
         if: ${{ contains(matrix.e2e-framework, 'cypress') }}
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -167,10 +167,10 @@ jobs:
 
       - name: Run e2e test script
         working-directory: ./playground
-        run: pnpm --filter '*${{ matrix.e2e-framework }}*' --workspace-concurrency 1 test:e2e
+        run: pnpm --filter '*${{ matrix.e2e-framework }}*' --filter '!*bare*' --workspace-concurrency 1 test:e2e
 
       - name: Cypress component testing for projects without Vitest
         if: ${{ contains(matrix.e2e-framework, 'cypress') }}
-        run: pnpm --filter '*cypress*' --filter '!*vitest*' --workspace-concurrency 1 test:unit
+        run: pnpm --filter '*cypress*' --filter '!*vitest*' --filter '!*bare*' --workspace-concurrency 1 test:unit
 
       # FIXME: `--with-tests` folders. It's failing now.
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -51,6 +51,7 @@ jobs:
           verification-script:
             - pnpm --filter '!*typescript*' build
             - pnpm --filter '*typescript*' build
+            # bare templates only contain vitest configurations, but do not include unit test files
             - pnpm --filter '*vitest*' --filter '!*bare*' test:unit
             - pnpm --filter '*eslint*' lint --no-fix --max-warnings=0
             - pnpm --filter '*prettier*' format --write --check
@@ -167,10 +168,12 @@ jobs:
 
       - name: Run e2e test script
         working-directory: ./playground
+        # bare templates can't pass e2e tests because their page structures don't match the example tests
         run: pnpm --filter '*${{ matrix.e2e-framework }}*' --filter '!*bare*' --workspace-concurrency 1 test:e2e
 
       - name: Cypress component testing for projects without Vitest
         if: ${{ contains(matrix.e2e-framework, 'cypress') }}
+        # bare templates only contain test configurations, but do not include component test files
         run: pnpm --filter '*cypress*' --filter '!*vitest*' --filter '!*bare*' --workspace-concurrency 1 test:unit
 
       # FIXME: `--with-tests` folders. It's failing now.
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -51,8 +51,7 @@ jobs:
           verification-script:
             - pnpm --filter '!*typescript*' build
             - pnpm --filter '*typescript*' build
-            # bare templates only contain vitest configurations, but do not include unit test files
-            - pnpm --filter '*vitest*' --filter '!*bare*' test:unit
+            - pnpm --filter '*vitest*' test:unit
             - pnpm --filter '*eslint*' lint --no-fix --max-warnings=0
             - pnpm --filter '*prettier*' format --write --check
             # FIXME: it's failing now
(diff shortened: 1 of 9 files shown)


cexbrayat:
I was leaning toward an option that would not be mutually exclusive of the others (and I think the linked issues were also asking for this). I would like to see a `--minimal` option that just does not generate the HelloWorld/WelcomeItem/TheWelcome/Icon* components but lets developers pick TS, vitest, etc.

Or maybe even change the default behavior to generate a lighter version.

haoqunjiang:
I think I can expand it to work with other options with some effort.

But what should we do when `pinia` and `router` is selected alongside `minimal`/`bare`? Do people really need a `router` template without any boilerplate code?

cexbrayat:
I think the complaint is mostly that the default setup is too verbose and the first thing you have to do after creating a new project is to delete all the components, all the CSS and assets etc.

Angular CLI for example only generates the `App` component, with an empty route config if the router is selected.
The Angular logo is inlined as SGV in the template which makes it easier to get rid of (even if the component's template is still way too verbose in my opinion).

That's why I would even consider simplifying the default instead of adding another option.
But I think adding another option without supporting the existing ones won't really answer what the community is asking in these issues IMHO.

haoqunjiang:
Refactored!

> That's why I would even consider simplifying the default instead of adding another option.

IMO we still need some kind of example code for new users, so the default is a bit more verbose than the ones in `create-vite`. In fact, I even included a composable example in the original implementation, only to be removed after the feedback by Evan… https://github.com/vuejs/create-vue/issues/2

I'm open to the idea of making the templates "easier to delete", but I need help on that.

runyasak:
> That's why I would even consider simplifying the default instead of adding another option.

I absolutely love and agree with the "easier to delete" approach, as mentioned in Evan's feedback.

In my opinion, adding one more prompt to ask about the bare configuration, with a sensible default choice, would be easier for new users who might not notice this feature in the docs. Additionally, we can provide a `--bare` argument for `create-vue` to skip the prompt and directly initialize with the "bare" option (set to `Yes` or `true`).

As an Angular user, I appreciate how the Angular CLI simplifies things with just prompts for the project name and stylesheet format, followed by generating a bare project. However, without some example components, it can be challenging for newcomers to understand the structure or workflow without reading the docs.

For my suggestion, I propose adding a `welcome` folder with easily removable example components. This would allow new users to follow our examples while keeping the project clean and straightforward for experienced developers. Alternatively, we could add a prompt, with an argument to skip it, to ensure users understand what they are configuring, enhancing the overall developer experience.

haoqunjiang:
> For my suggestion, I propose adding a `welcome` folder with easily removable example components.

I just tried this locally, and I found it made little difference compared to the current approach…
It's essentially moving the current `src/components` folder to `src/welcome`.

I also tried moving markups in `App.vue` into a standalone component inside `src/components` and leaving only one line of `import` and a single `<TheWelcome>` tag in `App.vue`. This did remove some visual clutter, but IMHO it's an inferior approach - we traded the opportunity to showcase essential Vue.js features for next-to-nothing, and the users still don't like it. Because according to the feedbacks in the linked issues, they want a plain `<p>` tag in `App.vue`, not even imports.

Anyway, even with all those experiments, I find the essential workflow in this PR still works, so I believe we can move forward with this implementation.

Now we need to decide	whether to add a user-facing prompt for this feature.
I like the word choice in #637 so I'm now inclined to add it here.

cexbrayat:
I have no stronger opinion about a user-facing prompt. If we go with it, I would use something other than "instructions", as it feels like you'll get a tutorial.

I haven't tested the PR, but I think we should add the `--bare` flag to the CI with a few examples (just bare, with TS, with tests, etc).

runyasak:
> For my suggestion, I propose adding a welcome folder with easily removable example components.

@haoqunjiang In my opinion, I would move the `welcome/` folder to `src/components`, making it `src/components/welcome/`. The `App.vue` file would remain in `src/` or its default position. This approach minimizes changes to the existing structure, keeping it familiar for beginners. It aligns with the `components` pattern, reducing potential confusion while maintaining consistency with the current folder structure.

> I have no stronger opinion about a user-facing prompt. If we go with it, I would use something other than "instructions," as it feels like you'll get a tutorial.
I haven't tested the PR, but I think we should add the `--bare` flag to the CI with a few examples (just bare, with TS, with tests, etc).

@cexbrayat I agree with using something other than "instructions" or adding more options in the tutorial.
Adding the `--bare` flag to the CI is a great idea, especially for experienced users who want to create a project without examples. We should also include documentation about the `--bare` flag in the README or docs so users can easily find and understand this option.

For now, I believe adding the `--bare` flag to `create-vue` is an excellent choice. It introduces a new configuration option without affecting the current user experience.

haoqunjiang:
> I haven't tested the PR, but I think we should add the `--bare` flag to the CI with a few examples (just bare, with TS, with tests, etc).

😂 I have to open another PR to refactor the CI first, because the job matrix would become way too complex after adding yet another flag…
https://github.com/vuejs/create-vue/pull/640

scripts/snapshot.mjs
```diff
@@ -85,10 +81,14 @@ for (const flags of flagCombinations) {
...
-    !featureFlagsDenylist.some((denylist) => denylist.every((flag) => combination.includes(flag))),
-)
+flagCombinations = flagCombinations
+  .filter(
+    (combination) =>
+      !featureFlagsDenylist.some((denylist) =>
+        denylist.every((flag) => combination.includes(flag)),
+      ),
+  )
+  .filter((combination) => !(combination.length === 1 && combination[0] === 'bare'))
```
cexbrayat, line 91: nit: can you add a comment here to explain why this line?

.github/workflows/ci.yml
```diff
@@ -49,13 +49,13 @@ jobs:
...
           os: [ubuntu-latest, windows-latest, macos-latest]
           verification-script:
-            - pnpm --filter "\!*typescript*" build
-            - pnpm --filter "*typescript*" build
-            - pnpm --filter "*vitest*" test:unit
-            - pnpm --filter "*eslint*" lint --no-fix --max-warnings=0
-            - pnpm --filter "*prettier*" format --write --check
+            - pnpm --filter '!*typescript*' build
+            - pnpm --filter '*typescript*' build
+            - pnpm --filter '*vitest*' --filter '!*bare*' test:unit
```
cexbrayat, line 54: Does that mean the bare flag doesn't work with the vitest flag?

scripts/snapshot.mjs
haoqunjiang: Added

.github/workflows/ci.yml
haoqunjiang: It works with all flags, but the test scripts would fail initially, because:
- it removes all test files for vitest, so vitest would fail by "No test files found".
- it simplified App.vue drastically, so the html structures won't match the example e2e tests' expectations.

haoqunjiang: I removed the example vitest specs because:
- they are used to test the components, but in the bare templates there isn't a `components` folder.
- the implementation is simpler with them removed.

I didn't touch the e2e specs because:
- I don't see people complaining about e2e specs being annoying
- I feel that people might still need some guidance on this. Because unlike unit tests, every e2e framework has different conventions on where to place the specs…

haoqunjiang: But we don't usually unit test our `App.vue`? I'm afraid it's not what users would expect.

cexbrayat: Our teams do (we unit test all our components). I think it would be better to have a unit test, even if it is elementary. This is also the approach taken by the Angular CLI for example.

haoqunjiang: I thought `App.vue` tests belong to E2E tests… Anyway, I've added it now

cexbrayat: ❤️ Looks great!

runyasak:
I just noticed the Vue template on [StackBlitz](https://stackblitz.com/). It includes only `App.vue` and `HelloWorld.vue` inside the `components/` folder, along with a single stylesheet file, `style.css`. This setup makes it very easy to remove example components and unnecessary configurations.

As a Vue user, I personally don’t see a need for both `style.css` and `main.css` initially. I haven’t encountered this issue, possibly because I almost always use Tailwind CSS or UnoCSS.

If we decide to go further with restructuring the folders and merging some files—similar to the [StackBlitz](https://stackblitz.com/) approach—it could make starting a project simpler and help users follow consistent patterns.

These are just my ideas.

image

haoqunjiang:
StackBlitz pulls from `create-vite` templates. Vite starter templates were [intentionally minimal](https://vite.dev/blog/announcing-vite3.html#create-vite-starter-templates). So they serve a different purpose than `create-vue` templates.
I agree that could be used as an inspiration for us to simplify the templates.

> As a Vue user, I personally don’t see a need for both `style.css` and `main.css` initially. I haven’t encountered this issue, possibly because I almost always use Tailwind CSS or UnoCSS.

We don't use `style.css`, but `base.css` along with `main.css`.
It's a common pattern among medium-to-large size projects. One is for CSS resets and common variables with minimum business logic and might be shared across projects, the other imports the base and includes actual global styles for the app.

If you are using Tailwind, you might have encountered it too, just in different forms: the `@tailwind base` directive is similar to `import 'base.css'` in plain CSS.
In the VitePress theme for https://vuejs.org, there're also both [`base.css`](https://github.com/vuejs/theme/blob/0c5e8a7d9ef6f7f23cc5694782c21ab629e3a5a6/src/core/styles/base.css) and [`index.css`](https://github.com/vuejs/theme/blob/0c5e8a7d9ef6f7f23cc5694782c21ab629e3a5a6/src/core/styles/index.css).
In Bootstrap, the names are a bit different; they use [`reboot.scss`](https://github.com/twbs/bootstrap/blob/0cbfe13adf669ad39ae9d8e873c2ad59befd3a3a/scss/bootstrap-reboot.scss) and [`bootstrap.scss`](https://github.com/twbs/bootstrap/blob/0cbfe13adf669ad39ae9d8e873c2ad59befd3a3a/scss/bootstrap.scss).

haoqunjiang:
Let's merge this PR now. Thanks for all the reviews!
I plan to ship another release just before the holidays, with one more PR to add documentations and a `--help` CLI option.

Regarding the remaining undecided matters like the `--bare` prompt and simplifying default templates, let's carry on the discussions in dedicated issues/PRs later.

PR #645, opened 2024-12-23
ci: add type-check step

cexbrayat:
### Description

#622 should fail on CI


haoqunjiang:
It's already failing because in TypeScript projects, `build` command also does type-checking: `run-p type-check \"build-only {@}\" --`

PR #646, opened 2024-12-23
feat: add `--help` and `--version` options

haoqunjiang:
### Description

Add documentation for the CLI flags. I think most of them are stable now.

I'm thinking about moving `--prettier` to a separate flag. But even in that case, `--eslint-with-prettier` can still be supported without much effort, so I choose to document it anyway.

### DIFF
--- a/index.ts
+++ b/index.ts
@@ -83,9 +83,9 @@ Available feature flags:
   --default
     Create a project with the default configuration without any additional features.
   --ts, --typescript
-    Add TypeScript Support.
+    Add TypeScript support.
   --jsx
-    Add JSX Support.
+    Add JSX support.
   --router, --vue-router
     Add Vue Router for SPA development.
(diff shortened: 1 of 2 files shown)


index.ts
```diff
@@ -61,33 +63,56 @@ function emptyDir(dir) {
...
+  --help
+    Display this help message.
+  --version
+    Display the version number of this CLI.
+
+Available feature flags:
+  --default
+    Create a project with the default configuration without any additional features.
+  --ts, --typescript
+    Add TypeScript Support.
```
cexbrayat, line 86: nit: I would lowercase "support" here and below

cexbrayat (approved):

index.ts
haoqunjiang: 😂 Well, I blame Copilot for the bad casing

PR #658, opened 2025-01-09
fix: correct the command to show help in README.md

shi0rik0:
The old command will show the help of `npm create`, not the help of `create-vue`.

### Description

Current documentation says we can use `npm create vue@latest -- --help` to show the help of `create-vue`, but I found it will print the help of `npm create`. I'm not sure if it's a bug of `npm`, but `npx create-vue@latest -- --help` works fine.

### My Enviroment

Windows 11 + npm 10.9.0

### DIFF
--- a/README.md
+++ b/README.md
@@ -19,6 +19,9 @@ npm create vue@latest
 
 By default the command will run in interactive mode, but you can also provide feature flags in the CLI arguments to skip the prompts. Run `npm create vue@latest -- --help` to see all available options.
 
+> [!NOTE]
+> If you're using PowerShell, you'll need to quote the `--`, that is, run `npm create vue@latest '--' --help`.
+
 If you need to support IE11, you can create a Vue 2 project with:
 
 ```sh


cexbrayat (changes requested):
Thanks for the PR

The `npm create vue@latest -- --help` command is correct on macOS and displays the help:

```
npm create vue@latest -- --help
Need to install the following packages:
create-vue@3.13.0
Ok to proceed? (y)


> npx
> create-vue --help

Usage: create-vue [FEATURE_FLAGS...] [OPTIONS...] [DIRECTORY]

Create a new Vue.js project.
Start the CLI in interactive mode when no FEATURE_FLAGS is provided, or if the DIRECTORY argument is not a valid package name.

Options:
  --force
    Create the project even if the directory is not empty.
  --bare
    Create a barebone project without example code.
  --help
    Display this help message.
  --version
    Display the version number of this CLI.
```

The `npx` version that works for you on Windows sadly doesn't work properly on macOS and executes create-vue as if a project was created:

```
npx create-vue@latest -- --help

Vue.js - The Progressive JavaScript Framework

✖ Add TypeScript ? … No / Yes
```

shi0rik0:
@cexbrayat I tried again and found that:
1. In `cmd`, the `npm` version works but the `npx` does not, jsut like on your MacOS.
2. In PowerShell, the `npx` version works but the `npm` does not.

This is interesting. I'll investigate it later.

shi0rik0:
@cexbrayat OK, I think I found out the answer. It's the PowerShell who is causing this problem behind the scene. According to my test and [this question](https://stackoverflow.com/a/15788023/25601920), the syntax of PowerShell will treat `--` specially, so executing `npm create vue@latest -- --help` will be treated as `npm create vue@latest --help`. A solution is to use `npm create vue@latest '--' --help`.

Do you think we should add a notice about this for Windows users?

shi0rik0:
@cexbrayat I've updated. Could you please review again?

cexbrayat (approved):
Thank you 👍

PR #695, opened 2025-02-18
feat!: use a giant multiselect instead of individual toggles for features

haoqunjiang:
As we are adding more features, the number of toggles is getting out of hand. A multiselect could save a few keystrokes for most users.

I also take this opportunity to change the prompts library to `@clack/prompts`.
(For context: `create-astro`, `sv`, `create-preact`, `create-solid`, `create-qwik` all use `@clack/prompts`, `create-vite` recently switched to `@clack/prompts` too, while `create-next-app` uses `prompts`)

The color library is changed to `picocolors`. IMO `kleur` is still the smaller library because of treeshaking. But `@clack/prompts` already comes with `picocolors` as a dependency, so by not adding `kleur`, we save a few bytes.

---

This is a proof-of-concept PR. I haven't put much consideration into the code style yet. Nor have I updated the i18n messages. I will do that if this change is accepted.

See the result in a screen recording:
[asciicast](https://asciinema.org/a/Aq334bHe0tJBpuxLNQyVAyzXb)
---

TODOs:
- [x] Clean up the code (wrap the `isCancel` condition; update the `needsEslint` type, etc.)
- [ ] I think we should remove the `packageName` related logic but it can be postponed into another PR
- [ ] Get the message translations reviewed

### DIFF
--- a/index.ts
+++ b/index.ts
@@ -4,7 +4,7 @@ import * as fs from 'node:fs'
 import * as path from 'node:path'
 
 import { parseArgs } from 'node:util'
-import { intro, outro, text, confirm, multiselect, select, isCancel } from '@clack/prompts'
+import { intro, outro, text, confirm, multiselect, select, isCancel, cancel } from '@clack/prompts'
 import { red, green, cyan, bold, dim } from 'picocolors'
 
 import ejs from 'ejs'
@@ -192,6 +192,15 @@ async function init() {
   const forceOverwrite = argv.force
 
   const language = getLanguage()
+  async function unwrapPrompt<T>(maybeCancelPromise: Promise<T | symbol>): Promise<T> {
+    const result = await maybeCancelPromise
+
+    if (isCancel(result)) {
+      cancel(red('✖') + ` ${language.errors.operationCancelled}`)
+      process.exit(0)
+    }
+    return result
+  }
 
   let result: {
     projectName?: string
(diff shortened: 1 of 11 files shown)


cexbrayat (approved):
Looks good, I like the idea 👍

locales/en-US.json
```diff
@@ -45,19 +45,19 @@
     }
   },
   "needsEslint": {
-    "message": "Add ESLint for code quality?",
+    "message": "Add ESLint for code quality",
```
btea, line 48: This is no longer a question. Can the `Yes` in the following options be removed? Otherwise, it looks a bit strange that both options contain `Yes`.

And the `negative` option has been removed, the `negative` attribute below can also be removed.

btea (approved):
LGTM

index.ts
```diff
@@ -166,261 +219,154 @@ async function init() {
...
+
+  if (!targetDir) {
+    targetDir =
+      result.projectName =
+      result.packageName =
+        await unwrapPrompt(
+          text({
+            message: language.projectName.message,
+            placeholder: defaultProjectName,
+            validate: (value) => (value.trim().length > 0 ? undefined : 'Should not be empty'),
```
cexbrayat, line 252: nit: this should be translated as well no? French translation: `Ne doit pas être vide`.

locales/fr-FR.json
```diff
@@ -13,51 +13,53 @@
     "message": "Nom du package\u00a0:",
     "invalidMessage": "Le nom du package.json est invalide"
   },
+  "featureSelection": {
+    "message": "Sélectionnez les fonctionnalités à inclure dans votre projet\u00a0:",
+    "hint": "(↑/↓ pour naviguer, espace pour sélectionner, a pour tout basculer, entrée pour confirmer)"
```
cexbrayat, line 18: `a pour tout basculer`-> `a pour tout sélectionner`

haoqunjiang:
Merging this PR now as I'm going to do another round of dependency maintenance and release a new version tomorrow.

PR #700, opened 2025-02-26
Update postversion.mjs

hiTommyhu:


hiTommyhu:
chore(postversion): use ‘const ’instead of ’let‘.

haoqunjiang (approved):

PR #701, opened 2025-03-01
fix: vitest merge config in form of callback

DingYuanhao:
I got an vitest error using `create-vue` template:

> Error: Vitest failed to start: 
> Error: Cannot merge config in form of callback
>   at mergeConfig (file://...)

Fixed it according to [vitest's official documentation](https://vitest.dev/config/#configuration)

image

### DIFF
--- a/template/config/vitest/vitest.config.js
+++ b/template/config/vitest/vitest.config.js
@@ -2,8 +2,8 @@ import { fileURLToPath } from 'node:url'
 import { mergeConfig, defineConfig, configDefaults } from 'vitest/config'
 import viteConfig from './vite.config'
 
-export default mergeConfig(
-  viteConfig,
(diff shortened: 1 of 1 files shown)


haoqunjiang:
But our vite config isn't defined as an function. Would you please open an issue with a runnable reproduction?

DingYuanhao:
> But our vite config isn't defined as an function. Would you please open an issue with a runnable reproduction?

Sorry, my mistake 😅 The issue lies with my `vite.config.ts` file, where I defined the Vite config in a functional way. I thought that defining the Vite config as mentioned in the PR would make it more universal. However, in this case, if it is not defined as a function, a new TypeScript error would be raised, even though Vitest runs successfully. Feel free to close the PR anyway :)

PR #709, opened 2025-03-09
chore(config): moving prettier script inside the script section

dennybiasiolli:
### Description

I noticed that after a project creation, there's a `format` outside of the scripts. That should be caused by the prettier configuration script.

This PR solves that small issue.

### DIFF
--- a/template/config/prettier/package.json
+++ b/template/config/prettier/package.json
@@ -1,5 +1,7 @@
 {
-  "format": "prettier --write src/",
+  "scripts": {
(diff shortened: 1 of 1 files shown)


cexbrayat (approved):
Good catch, thanks for the PR @dennybiasiolli

cexbrayat:
@haoqunjiang You may want to do a patch release with this?

haoqunjiang:
Oops, what an oversight. Thank you!!

PR #760, opened 2025-06-09
feat: add `rolldown-vite` features

### LINKED ISSUE #756: New experiemental flag: Use Rolldown-Vite
Now that [it's officially announced](https://voidzero.dev/posts/announcing-rolldown-vite), I think we should provide an option for users to try it out in their projects.

I think we can group the `rolldown-vite` option along with the `oxlint` option in a multi-select as "experimental features", and we can reuse that list in the future for other things, like `tsgo` integration, etc.

btea:
### Description
close #756

### DIFF
--- a/index.ts
+++ b/index.ts
@@ -39,8 +39,9 @@ const FEATURE_FLAGS = [
   'playwright',
   'eslint',
   'prettier',
-  'eslint-with-oxlint',
   'eslint-with-prettier',
+  'oxlint',
+  'rolldown-vite',
 ] as const
 
 const FEATURE_OPTIONS = [
@@ -76,6 +77,20 @@ const FEATURE_OPTIONS = [
     value: 'prettier',
     label: language.needsPrettier.message,
   },
+  {
+    value: 'experimental features',
+    label: language.needExperimenttal.message,
+  },
+] as const
+const EXPERIMENTAL_FEATURE_OPTIONS = [
+  {
+    value: 'oxlint',
+    label: language.needsOxlint.message,
+  },
+  {
+    value: 'rolldown-vite',
+    label: language.needsRolldownVite.message,
+  },
 ] as const
 
 type PromptResult = {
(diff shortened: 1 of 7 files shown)


cexbrayat (approved):

locales/fr-FR.json
```diff
@@ -60,7 +60,17 @@
...
   },
   "needsOxlint": {
-    "message": "Installer Oxlint pour un linting plus rapide\u00a0? (expérimental)"
+    "message": "Oxlint (expérimental)"
+  },
+  "needExperimenttal": {
+    "message": "Activer les fonctionnalités expérimentales"
+  },
+  "needsExperimentalFeatures": {
+    "message": "Sélectionnez les fonctionnalités expérimentales à inclure:",
```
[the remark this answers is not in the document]
btea: Done.

index.ts
```diff
@@ -76,6 +77,20 @@ const FEATURE_OPTIONS = [
     value: 'prettier',
     label: language.needsPrettier.message,
   },
+  {
+    value: 'experimental features',
```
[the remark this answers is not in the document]
btea: looks great.

index.ts
```diff
@@ -76,6 +77,20 @@ const FEATURE_OPTIONS = [
     value: 'prettier',
     label: language.needsPrettier.message,
   },
+  {
+    value: 'experimental-features',
+    label: language.needsExperimental.message,
```
haoqunjiang, line 82: I prefer not to make "Enable experimental features" a feature selection.
Just looking at this description will not provide users with enough knowledge about the possible experimental features to make an informed decision.

So, let's just ask all the users about the experimental features. As long as the default is none, it won't disrupt the workflow.

haoqunjiang (changes requested):

index.ts
btea: Does oxlint keep the original logic?

haoqunjiang: Sorry for failing to express it clearly in the first place 😂 Had a hard time to come up with the correct words.

PR #761, opened 2025-06-13
fix: target directory setting default value

btea:
### Description
Hopefully I can press enter to quickly select the default as the target directory.

### DIFF
--- a/index.ts
+++ b/index.ts
@@ -260,9 +260,13 @@ async function init() {
         message: language.projectName.message,
         placeholder: defaultProjectName,
         defaultValue: defaultProjectName,
+        validate: (value) =>
+          value.length === 0 || value.trim().length > 0
+            ? undefined
+            : language.projectName.invalidMessage,
       }),
     )
-    targetDir = result.projectName = result.packageName = _result.trim() || defaultProjectName
+    targetDir = result.projectName = result.packageName = _result.trim()
   }
 
   if (!canSkipEmptying(targetDir) && !forceOverwrite) {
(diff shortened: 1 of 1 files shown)


index.ts
```diff
@@ -259,11 +259,10 @@ async function init() {
       text({
         message: language.projectName.message,
         placeholder: defaultProjectName,
-        validate: (value) =>
-          value.trim().length > 0 ? undefined : language.projectName.invalidMessage,
```
haoqunjiang, line 263: Don't remove the `validate` function, we still need it to reject inputs with whitespaces only, like `  `.

haoqunjiang: I'm okay with converting space-only inputs to `vue-project`, but the problem with the implementation is that the feedback isn't as clear:
Hit Enter directly, `vue-project` is displayed as the answer to the prompt:
image

Use two spaces, we still use `vue-project` as the project name but the answer on screen is empty:
image

haoqunjiang: The easiest fix I found is to accept `''` but reject `'  '`.
But feel free to propose any other way to work around this issue.

btea: That makes sense. At the moment, `The easiest fix I found is to accept '' but reject '  '` seems to be the simplest and most reasonable modification. 👍

btea:
I don't understand why it would cause the ci to fail.🤯

haoqunjiang:
😂 It's fine. The CI broke because I used `no-frozen-lockfile`, and the new ESLint version broke the `typescript-eslint` type. It will be fixed tomorrow with a new release from `typescript-eslint`.

PR #775, opened 2025-07-01
chore: base template align node version with vite 7

btea:
### Description

sync #774

### DIFF
--- a/template/base/package.json
+++ b/template/base/package.json
@@ -1,6 +1,9 @@
 {
   "private": true,
   "type": "module",
+  "engines": {
+    "node": "^20.19.0 || >=22.12.0"
+  },
   "scripts": {
     "dev": "vite",
     "build": "vite build",


haoqunjiang:
I think `npm install` and similar commands would already throw on unsupported Node versions when the package manager tries to install vite? If so, we don't need to add this to the template explicitly. It doesn't add any new information to the project, yet it may become a maintenance burden in the future.

btea:
I now use Node 18 to run `pnpm create vue` to create a project and execute `pnpm i`, and no errors or prompts appear. It is not until `pnpm dev` is executed that the error appears.

Therefore, if the required node version information is added to package.json, the user can be informed more clearly of the reason.

In fact, after Vite released 7, many people encountered this problem and raised issues.

haoqunjiang (approved):

PR #813, opened 2025-09-02
chore: remove experimental from oxlint

TheAlexLichter:
### Description

A rather "quick fix" as [Oxlint is stable by now](https://voidzero.dev/posts/announcing-oxlint-1-stable). Vue support is still focused on the `<script>` blog only so leaving it in the experimental selection seems fine for me.

### DIFF
--- a/locales/fr-FR.json
+++ b/locales/fr-FR.json
@@ -60,7 +60,7 @@
     }
   },
   "needsOxlint": {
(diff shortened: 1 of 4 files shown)


cexbrayat (changes requested):
Thanks for the PR @TheAlexLichter 

Would you mind updating the other locales as well please?

PR #814, opened 2025-09-05
fix: Correct "build" command in package.json

Abilovv599:
some d.ts files will generate after build. type-check fails when those files aren't generated. That is why we should build first.


Abilovv599:
Hi, I’m using Nuxt UI, and I noticed that auto-imports.d.ts and components.d.ts are generated at build time. The documentation recommends adding these files to .gitignore. However, I couldn’t reproduce the same issue locally. I think the difference is that in the real repository, I have many more components, so it takes longer to generate those files.

PR #824, opened 2025-09-20
feat: add `needsAutomaticalllyInstallDependencies` option

### LINKED ISSUE #822: Add an option to automatically install dependencies.
### Please describe
After the project initialization is completed, add the option of whether to automatically install dependencies. If the user selects true, the dependencies will be installed directly.

### Additional context
Because the project we initialize must install dependencies before it can run, if you can directly choose automatic installation, you can omit the manual installation command operation.

btea:
### Description
close #822


btea:
https://github.com/vuejs/create-vue/issues/822#issuecomment-3317579774

PR #869, opened 2025-12-12
feat: always include Oxlint with ESLint

### LINKED ISSUE #886: Choosing Oxlint without ESLint does not add Oxlint
### Describe the bug

When choosing only Oxlint without selecting ESLint, it won't install Oxlint at all.

### Expected behavior

Oxlint being installed.

### How to reproduce

Use the following CLI choices to reproduce

```
◇  Project name (target directory):
│  vue-project
│
◇  Select features to include in your project: (↑/↓ to navigate, space to select, a to toggle all, enter to
confirm)
│  TypeScript, Router (SPA development)
│
◇  Select experimental features to include in your project: (↑/↓ to navigate, space to select, a to toggle
all, enter to confirm)
│  Oxlint, rolldown-vite (experimental)
│
◇  Skip all example code and start with a blank Vue project?
│  No
```

Image

haoqunjiang:
Oxlint has been stable for half a year now.
I think it's time to make it the default for all linter users.

Also closes #886

### DIFF
--- a/index.ts
+++ b/index.ts
@@ -158,7 +158,7 @@ const helpMessage = `\
 Usage: create-vue [FEATURE_FLAGS...] [OPTIONS...] [DIRECTORY]
 
 Create a new Vue.js project.
-Start the CLI in interactive mode when no FEATURE_FLAGS is provided, or if the DIRECTORY argument is not a valid package name.
+Runs in interactive mode if started without feature flags, or if DIRECTORY is missing or not a valid package name.
 
 Options:
   --force


haoqunjiang:
I opened https://github.com/vuejs/create-vue/pull/875 to simplify the generation process (it's really hard to make changes with the current implementation) and added `.oxlintrc.json` in the generated projects.

I had thought oxlint should be the default zero-config linting tool, as it's so fast and prioritized `correctness` rules in its early development.
But yeah, it's a bit hard to ignore the inherent complexity now. Yet it only supports JSON configurations, so I don't have a good idea to fix that.

As for TypeScript, I've tried my best to encapsulate the complexities into `@vue/tsconfig`, but the generated tsconfigs haven't been documented well, which could be improved before making TS the default language.

PR #885, opened 2025-12-31
feat: oxlint config add `ignorePatterns`

btea:
### Description
It appears that oxlint does not exclude node_modules by default.


TheAlexLichter:
@btea update: this has been changed in Oxlint v1.36.0. `.gitignore` his always honored by default now. https://github.com/oxc-project/oxc/pull/17375

PR #1010, opened 2026-05-11
fix: quote glob in package script

### LINKED ISSUE #951: Linter is not working when used with bun on Windows
### Describe the bug
I am trying to lint using oxfmt with `bun`, it is giving me below error

Image

### Expected behavior

It should lint it properly without any errors whatsoever

### How to reproduce

Create a new vuejs project using bun with oxfmt. And after that try running the `lint` script as mentioned below. The below scripts are autocreated when a new project is created using bun and oxfmt. It also adds `oxlint` script for including eslint. It is working with `npm` but not with `bun`.

```json
{
    "lint": "run-s lint:*",
    "lint:oxlint": "oxlint . --fix",
    "lint:eslint": "eslint . --fix --cache",
}
```

the command would be 
```bash
bun run lint
```

clemyan:
### Description

When using oxlint, the `lint` script in the generated `package.json` contains a `*` that may be expanded by the shell and can lead to unexpected behavior

This PR wraps the arg in quotes to stop shells from expanding it as a glob

This should fix #951, but I can't confirm right now

### DIFF
--- a/template/linting/oxlint/package.json
+++ b/template/linting/oxlint/package.json
@@ -2,7 +2,7 @@
   "scripts": {
     "lint:oxlint": "oxlint . --fix",
     "lint:eslint": "eslint . --fix --cache",
-    "lint": "run-s lint:*"
+    "lint": "run-s 'lint:*'"
   },
   "devDependencies": {
     "eslint-plugin-oxlint": "~1.63.0",
--- a/template/linting/oxlint/package.json
+++ b/template/linting/oxlint/package.json
@@ -2,7 +2,7 @@
   "scripts": {
     "lint:oxlint": "oxlint . --fix",
     "lint:eslint": "eslint . --fix --cache",
-    "lint": "run-s 'lint:*'"
+    "lint": "run-s \"lint:*\""
   },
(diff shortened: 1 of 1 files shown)


template/linting/oxlint/package.json
```diff
@@ -2,7 +2,7 @@
   "scripts": {
     "lint:oxlint": "oxlint . --fix",
     "lint:eslint": "eslint . --fix --cache",
-    "lint": "run-s lint:*"
+    "lint": "run-s 'lint:*'"
```
skirtles-code: Single quotes don't work reliably in all shells. For example, this didn't work for me when testing on Windows with the built-in Command Prompt (I was testing using pnpm, not bun). The quotes are treated as part of the value, so it tries to run a task called `'lint:*'`, which doesn't exist.
```suggestion
    "lint": "run-s \"lint:*\""
```

skirtles-code (approved):
Looks good.

This fix is also needed for Yarn, not just bun.

haoqunjiang (approved):
Thanks! And thanks @skirtles-code for the review!

PR #1011, opened 2026-05-11
fix: add missing dependency

clemyan:
### Description

The `eslint-plugin-vue` has a peer dependency on `vue-eslint-parser` but the generated `package.json` does not list it as a dependency.

I have added the `vue-eslint-parser` as a devDependency in the template.


skirtles-code:
To provide a little more background on the underlying issue...

I believe this change is intended to fix linting with Yarn. I could reproduce the problem fairly readily with the latest Yarn (4.15.0) and a `create-vue` project that had linting enabled. I had to use `yarn lint:eslint` to run the linting, as `yarn lint` fails for unrelated reasons.

I posted a full reproduction here:

- https://github.com/skirtles-code/vue-yarn-repro

That reproduction is using workspaces, though they aren't needed to reproduce this issue with `create-vue`. I included workspaces because they were needed to reproduce a separate problem with Yarn and Vue DevTools.

haoqunjiang (approved):

PR #1013, opened 2026-05-12
chore(ci): set explicit least-privilege workflow permissions

arpitjain099:
## Summary
- add an explicit `permissions` block to the main CI workflow
- set `contents: read` to follow least-privilege defaults for `GITHUB_TOKEN`
- keep current CI behavior unchanged while reducing token scope risk

## Why
GitHub Actions defaults can be broader than needed. Making permissions explicit improves security posture and keeps future permission changes intentional and reviewable.


haoqunjiang (approved):
Thanks!

PR #1065, opened 2026-06-30
chore: enable tsconfig `allowArbitraryExtensions`

btea:
### Description

refer to https://github.com/vitejs/vite/issues/22772


cexbrayat (commented):
I personally think that WASM support is a bit too specific to be added to every Vue project. I would rather keep the tsconfig as light as possible.

btea:
Yeah, you're right. Wasm is indeed unrelated to Vue, and this is a Vite issue, which they've already documented. Let's leave it as it is.

PR #1066, opened 2026-07-01
feat: detect nub package manager

colinhacks:
When invoked through nub, create-vue misdetects the package manager as npm. It infers the PM from `npm_config_user_agent` in `utils/packageManager.ts` and falls back to npm; [nub](https://nubjs.com) advertises `nub/<version>`, so it isn't matched.

nub is a Rust CLI with a pnpm-compatible command grammar. This adds it to `inferPackageManager` and the options list, and routes it through the `run`-prefixed form in `getCommand` — nub requires an explicit `nub run <script>` and has no implicit script shortcut — yielding `nub install` and `nub run dev`.


colinhacks:
Just checking in on this one — ready whenever you get a chance to take a look.

haoqunjiang (approved):
Thanks! LGTM.

PR #1100, opened 2026-08-04
fix: clarify that the overwrite prompt deletes all existing files

### LINKED ISSUE #1099: UX/CLI: Ambiguous phrasing "Remove existing files and continue" leads to accidental file deletion
### Context
When running `npm create vue@latest` inside an existing project directory containing non-Vue files, the CLI presents a selection prompt regarding the non-empty directory.

#### The Problem
One of the selectable options is phrased as:
`Remove existing files and continue`

Because this uses imperative phrasing without explicit indication that the tool itself will immediately delete the directory contents, it can easily be misread as a pre-requisite instruction for the user (i.e., "Please manually remove your existing files before continuing").

Selecting this option without realizing it is an immediate emptyDir() action wiped the entire host project (including untracked files like `.env`).

### Suggested Improvement
Make the copy explicitly destructive or warning-driven, for example:

- [DANGER] Delete all existing files in this directory and continue

- Wipe current directory contents and overwrite

- Or add a secondary explicit confirmation prompt before performing the destructive filesystem operation: "Are you sure you want to permanently delete X files in this directory? (y/N)"

- At least a `?` sign could be helpful.

lazerg:
### Description

The confirm prompt for a non-empty target directory reads "Remove existing files and continue". Without a question mark it looks like an instruction telling the user to clear the folder first, not a question about what the CLI is about to do. In #1099 someone answered Yes on that reading and lost the contents of an existing project, including untracked files like `.env`.

This rewords it to "Remove all existing files and continue?", so it reads as a question and states that everything in the directory goes. The other locales already phrase it as a question.

Fixes #1099

### DIFF
--- a/locales/en-US.json
+++ b/locales/en-US.json
@@ -8,7 +8,7 @@
       "current": "Current directory",
       "target": "Target directory"
     },
-    "message": "is not empty. Remove existing files and continue"
+    "message": "is not empty. Remove all existing files and continue?"
   },
   "packageName": {
     "message": "Package name:",


cexbrayat (approved):
Looks like the question mark was lost in https://github.com/vuejs/create-vue/commit/3aaddc6b4ef40fe1291cc5ce2d16c7bdbb5a6b1f Thanks for the fix

PR #1144, opened 2026-09-17
fix: handle clack cancel symbol in prompt wrapper

btea:
### Description

image

https://github.com/bombshell-dev/clack/pull/601


cexbrayat (approved):
Thanks!

